PR #259, opened 2021-06-21
dependency updates, use npm workspaces

jescalan:
Updates all dependencies, switches over to npm workspaces, works out some weird peer dep situations.


packages/code-block/utils/prism-utils/index.test.js
```diff
@@ -9,7 +9,9 @@ const syntaxesList = Object.keys(syntaxes).reduce((acc, slug) => {
   return acc.concat(slug, aliases)
 }, [])
 
-describe('syntaxes', () => {
+// TODO: I'm not sure if we want to always run these tests, it seems like the list of syntaxes we provide should align with the needs of our
+// content, not necessarily all of the possible syntaxes exposed by the library.
+describe.skip('syntaxes', () => {
```
BRKalow, line 14: @zchsh I'm offering this up for discussion. The upgrade to refractor introduced a bunch of new syntaxes, most of which seem pretty obscure. I'm thinking it might make more sense to update our mapping as our authors want to utilize new syntaxes, as opposed to doing it whenever we upgrade the underlying lib. Happy to continue with the current approach if we think it's a better way, though!

next.config.js
```diff
@@ -1,33 +1,18 @@
...
 })(
   withSwingset({
     componentsRoot: 'packages/*',
     docsRoot: 'docs/*',
   })({
+    eslint: {
+      // Warning: Dangerously allow production builds to successfully complete even if
+      // your project has ESLint errors.
+      ignoreDuringBuilds: true,
+    },
```
BRKalow, line 15: Ticket to add this to our next plugin: https://app.asana.com/0/1100423001970639/1200795708532158/f

package.json
```diff
@@ -3,69 +3,70 @@
...
+    "npm": ">=7.0.0",
+    "node": ">=10.0.0"
+  },
+  "resolutions": {
+    "react": "17.0.2",
+    "react-dom": "17.0.2"
   },
   "scripts": {
-    "bootstrap": "lerna bootstrap -- --no-optional",
-    "bs": "npm run bootstrap",
```
BRKalow, line 45: Removing references to `lerna bootstrap` as everything is handled via `npm` now.

packages/logo-grid/style.css
```diff
@@ -1,11 +1,12 @@
-@import '~tippy.js/themes/light.css';
+@import '~@tippyjs/react/node_modules/tippy.js/dist/tippy.css';
```
BRKalow, line 1: `tippy.js` is not getting hoisted now, so need to reference it as nested.

packages/sentinel-embedded/index.test.js
```diff
@@ -1,4 +1,5 @@
 import { render, waitFor } from '@testing-library/react'
+import { setImmediate } from 'timers'
```
BRKalow, line 2: `setImmediate` is no longer polyfilled by jest, so importing the native one explicitly.

packages/tabs/index.js
```diff
@@ -50,6 +42,14 @@ function Tabs({
     }
   }, [])
 
+  if (!children) {
+    process.env.NODE_ENV !== 'production' &&
+      console.warn(
+        '@hashicorp/react-tabs: There are no `Tab` children for the `Tabs` component to render.'
+      )
+    return null
+  }
```
BRKalow, line 51: Moved to avoid a conditional hooks call lint error

packages/text-split/partials/checkbox-list/index.js
```diff
@@ -6,8 +6,8 @@ import CheckSquare from './icons/check-square.svg?include'
 function CheckboxList({ items, theme, product }) {
   const validItems = items && items.filter((l) => l !== '')
   const hasItems = validItems && validItems.length > 0
-  if (!hasItems) return null
   const { themeClass } = useProductMeta(product) // overrides base --brand with product colors
+  if (!hasItems) return null
```
BRKalow, line 10: Addressing a conditional hooks call error

packages/code-block/utils/prism-utils/index.test.js
zchsh: Very on board with this - very much agreed we should focus on what syntaxes are actually in use. [Made a task for the platform board](https://app.asana.com/0/1100423001970639/1200801692661223/f) to assess what our new list of supported syntaxes would be 👍

zchsh (approved):
Looks good - I pulled down the branch and ran things locally 👍 Install seems quite a bit faster than with `lerna bootstrap`! Excited for `changesets` 💖

packages/logo-grid/style.css
jescalan: I'm worried about this - if it is hoisted in a host installation doesn't this break?

BRKalow: Yeah I thought about that as well...let me do some quick validation here.

BRKalow: So far this has been working just fine. As we are no longer depending directly on `tippy.js`, it should always be nested in `@tippyjs/react`. Tested on www with npm v6 and v7.

I agree it's not ideal, but unfortunately I can't think of a better alternative.

jescalan: Yeah I can't really either. The issue I anticipate is if a project uses two different packages that both have tippy as a dep and it decides to hoist them. I guess we can just deal with this if/when it happens. Setting the version ranges using a carat seems to help to prevent things from being hoisted in this way a little.

PR #273, opened 2021-07-22
fix(terminal): ensure pre has same bg color

BRKalow:
🎟️ [Asana Task]()
🔍 [Preview Link](https://react-components-git-brkfix-command-line-termi-0508a5-hashicorp.vercel.app/components/commandlineterminal)

---

## Description

Noticed when updating waypoint that the pre inside of the terminal had a different background color. I updated the styles of the terminal component to set the pre background explicitly. Not sure if there's a better way to do this 🤔. @zchsh curious what you think of this in particular.

Regression:
Screen Shot 2021-07-22 at 11 56 50 AM

Also visible here: https://react-components.vercel.app/components/commandlineterminal

## Release Information

Please select one (**required**)

- [ ] **Major** (This PR introduces a breaking (incompatible API) change)
- [ ] **Minor** (This PR adds functionality but it backwards-compatible)
- [x] **Patch** (This PR adds a backwards-compatible bug fix)

<details>
<summary>Release Notes (<strong>required</strong>)</summary>

</details>

---

### PR Checklist 🚀

- [ ] Write a useful description (above) to give reviewers appropriate context.
- [ ] Add release notes to the appropriate section (above).
- [ ] Conduct reasonable cross browser testing for both compatibility and responsive behavior (We have a [Cross Browser Testing](https://crossbrowsertesting.com) account for this, if you don't have access, just ask!).

### DIFF
--- a/packages/terminal/Terminal.module.css
+++ b/packages/terminal/Terminal.module.css
@@ -50,7 +50,7 @@
     border-radius: 0 0 var(--terminal-radius) var(--terminal-radius);
     padding: 24px 32px;
 
-    & > pre {
+    & pre {
       background: var(--terminal-body-color);
     }
   }
--- a/packages/terminal/Terminal.module.css
+++ b/packages/terminal/Terminal.module.css
@@ -50,6 +50,7 @@
     border-radius: 0 0 var(--terminal-radius) var(--terminal-radius);
     padding: 24px 32px;
 
+    /* Ensures that highlighted code maintains the same background as the terminal element */
     & pre {
       background: var(--terminal-body-color);
     }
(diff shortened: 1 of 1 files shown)


zchsh:
@BRKalow I _think_ this will be fixed once https://github.com/hashicorp/mktg-global-styles/pull/9/files#r674060211 lands, as it will remove the globally scoped rule on `pre` that I think might be setting the mismatched background colour. The removal of that global style should give us the flexibility to set the `background-color` if we want, or leave it the default transparent / `none` as it seems we'd ideally want to do here.

zchsh (commented):
@BRKalow I _think_ this will be fixed once https://github.com/hashicorp/mktg-global-styles/pull/9/files#r674060211 lands, as it will remove the globally scoped rule on `pre` that I think might be setting the mismatched background colour. The removal of that global style should give us the flexibility to set the `background-color` if we want, or leave it the default transparent / `none`. Latter seems like the ideal solution in this specific case IMO.

PR #276, opened 2021-07-28
Show Consent Manager globally

mwickett:
📝 [RFC](https://docs.google.com/document/d/12UIDBgpR6lilkFrxQFr7i1zTxCm04szAZkGKSSlgRkI/edit#)

---

## Description

This removes the `@segment/in-eu` package and check from our Consent Manager. This change means it will now be shown to all visitors, regardless of their geography. 

Also updates tests, and fixes some small bugs (I think..):

* The `forceShow` prop wasn't working quite right because of the wrong conditional order in `shouldRequireConsent`
* The `preference` object key names were off (e.g. `All` instead of `loadAll`) - prime example of where a TS refactor would help us out here, but that's outside my scope at the moment unfortunately)

### PR Checklist 🚀

      1

### DIFF
--- a/packages/consent-manager/docs.mdx
+++ b/packages/consent-manager/docs.mdx
@@ -5,6 +5,7 @@ componentName: ConsentManager
 A GDPR-compliant consent manager, integrated with segment.
 
 <LiveComponent>{`<ConsentManager
+  version={1}
   segmentWriteKey='iyi06c432UL7SB1r3fQReec4bNwFyzkW'
   utilServerRoot='https://hashicorp-web-util-staging.herokuapp.com'
   privacyPolicyLink='https://www.hashicorp.com/privacy'
@@ -28,7 +29,7 @@ A GDPR-compliant consent manager, integrated with segment.
       url: 'http://www.an-optional-url-for-a-script-to-add-to-the-page.com',
     },
     {
-      name: 'Name of the service',
+      name: 'Name of the service 2',
       category: 'Example Category',
       description: 'A script with additional elements to be injected',
       body: '',
--- a/packages/consent-manager/index.js
+++ b/packages/consent-manager/index.js
@@ -1,6 +1,5 @@
 import { Component } from 'react'
 import EventEmitter from 'events'
-import inEU from '@segment/in-eu'
 import ConsentBanner from './partials/banner'
 import ConsentPreferences from './partials/dialog'
 import { loadPreferences, savePreferences } from './partials/cookies'
@@ -47,7 +46,11 @@ export default class ConsentManager extends Component {
   }
 
   shouldRequireConsent() {
-    // 1. Check cookies
+    // 1. If prop override is set, always show the consent bar.
+    if (this.props.forceShow) {
+      return true
+    }
+    // 2. Check cookies and apply existing consent preferences if they've already been set.
     if (
       this.state.preferences &&
       this.state.preferences.version === this.props.version
@@ -60,14 +63,8 @@ export default class ConsentManager extends Component {
       )
       return false
     }
-    // 2. Check inEU. If in-eu, show consent. If not, show nothing. Load all.
-    if (inEU() || this.props.forceShow) {
-      return true
-    } else {
-      // Load all analytics
-      this.saveAndLoadAnalytics({ loadAll: true })
-      return false
-    }
+    // 3. Always show the consent bar if we don't have a response stored.
+    return true
   }
 
   componentDidMount() {
--- a/packages/consent-manager/index.js
+++ b/packages/consent-manager/index.js
@@ -26,7 +26,7 @@ export default class ConsentManager extends Component {
 
   saveAndLoadAnalytics(preferences) {
     if (typeof preferences === 'undefined') {
-      preferences = { All: false, 'Segment.io': false }
+      preferences = { loadAll: false, segment: false }
     }
     savePreferences(preferences, this.props.version)
 
@@ -52,7 +52,7 @@ export default class ConsentManager extends Component {
     }
     // 2. Check cookies and apply existing consent preferences if they've already been set.
     if (
-      this.state.preferences &&
+      Object.keys(this.state.preferences).length !== 0 &&
       this.state.preferences.version === this.props.version
     ) {
       // Load segment and custom integrations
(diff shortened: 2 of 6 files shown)


packages/consent-manager/index.js
```diff
@@ -47,7 +46,11 @@ export default class ConsentManager extends Component {
   }
 
   shouldRequireConsent() {
-    // 1. Check cookies
+    // 1. If prop override is set, always show the consent bar.
+    if (this.props.forceShow) {
+      return true
+    }
+    // 2. Check cookies and apply existing consent preferences if they've already been set.
```
mwickett, line 53: Moved this check to the top - I actually believe this was a bug that caused the prop not to work if you'd set consent preferences previously.

packages/consent-manager/props.js
```diff
@@ -21,6 +21,11 @@ module.exports = {
     description:
       "root path of the instance of HashiCorp's `web-utility-server` to use. This is used to fetch integrations based on segment write key",
   },
+  forceShow: {
+    type: 'boolean',
+    description:
+      'Useful for development, setting this to true will always show the consent bar even if a cookie with existing preferences is stored',
+  },
```
mwickett, line 28: This wasn't documented before, but it is used in the live code Swingset example.

BRKalow (approved):
Looks reasonable to me!

BRKalow:
@mwickett Looks like we need to update some tests here 👀

mwickett:
🤦🏻‍♂️ Absolutely. I skipped a step there. I'll update those and re-tag you for a look.

packages/consent-manager/docs.mdx
```diff
@@ -5,6 +5,7 @@ componentName: ConsentManager
 A GDPR-compliant consent manager, integrated with segment.
 
 <LiveComponent>{`<ConsentManager
+  version={1}
```
mwickett, line 8: `version` is used to check if consent needs to be requested again.

packages/consent-manager/docs.mdx
```diff
@@ -28,7 +29,7 @@ A GDPR-compliant consent manager, integrated with segment.
       url: 'http://www.an-optional-url-for-a-script-to-add-to-the-page.com',
     },
     {
-      name: 'Name of the service',
+      name: 'Name of the service 2',
```
mwickett, line 32: Name is used as a key, so the duplication here was throwing a react warning.

packages/consent-manager/index.js
```diff
@@ -27,7 +26,7 @@ export default class ConsentManager extends Component {
 
   saveAndLoadAnalytics(preferences) {
     if (typeof preferences === 'undefined') {
-      preferences = { All: false, 'Segment.io': false }
+      preferences = { loadAll: false, segment: false }
```
mwickett, line 29: You'll see this change here and in the tests - somehow the property names had drifted and were incorrect. The [`loadAnalytics` function](https://github.com/hashicorp/react-components/blob/8dfad01e23f24496f73952d3b7fc6b7c5f3c725c/packages/consent-manager/partials/load.js#L20-L33) (in `partials/load`) is looking for `loadAll`, `segment` and `custom` so have adjusted these appropriately here.

packages/consent-manager/index.js
```diff
@@ -59,14 +62,9 @@ export default class ConsentManager extends Component {
...
-    }
-    // 2. Check inEU. If in-eu, show consent. If not, show nothing. Load all.
-    if (inEU() || this.props.forceShow) {
-      return true
     } else {
-      // Load all analytics
-      this.saveAndLoadAnalytics({ loadAll: true })
-      return false
+      // 3. Always show the consent bar if we don't have a response stored.
+      return true
```
mwickett, line 67: Tucked this return inside the conditional just to make it clearer.

packages/consent-manager/index.test.js
```diff
@@ -49,42 +49,19 @@ const defaultProps = {
     },
   ],
   container: '#consent-manager',
-  forceShow: true,
```
mwickett, line 52: `forceShow` is a "dev only" prop, so having it be in default props here didn't make sense to me. It's manually set on almost every test anyways.

packages/consent-manager/index.test.js
```diff
@@ -49,42 +49,19 @@ const defaultProps = {
...
-  await runWithMockedImport(
-    '@segment/in-eu',
-    () => false,
-    (MockedConsentManager) => {
-      render(<MockedConsentManager {...defaultProps} forceShow={false} />)
-      expect(screen.queryByTestId('consent-banner')).not.toBeInTheDocument()
-    }
-  )
-})
-
```
mwickett, line 81: No more EU check, so 🔪

packages/consent-manager/index.test.js
```diff
@@ -49,42 +49,19 @@ const defaultProps = {
...
 test('sets existing preferences and does not show the banner if preferences are already set', async () => {
   await runWithMockedImport(
     './partials/cookies',
     {
       loadPreferences: () => {
-        return { All: false, 'Segment.io': false }
+        return { loadAll: false, segment: false, version: 0 }
```
mwickett, line 64: The missing version was causing this test to fail, and I updated the preference keys per my above comment.

BRKalow (approved):
Solid cleanup and comments, LGTM!

packages/consent-manager/index.js
```diff
@@ -59,14 +62,9 @@ export default class ConsentManager extends Component {
         this.props.additionalServices
       )
       return false
-    }
-    // 2. Check inEU. If in-eu, show consent. If not, show nothing. Load all.
-    if (inEU() || this.props.forceShow) {
-      return true
     } else {
-      // Load all analytics
-      this.saveAndLoadAnalytics({ loadAll: true })
```
jescalan, line 68: Just want to make sure removing this still would make it work?

jescalan (approved):
There's one line deletion that I am 👀 about, but otherwise this looks great! Thanks for patching up a bunch of other stuff too.

packages/consent-manager/index.js
mwickett: we're good here - this line was the "not in the EU, so load everything by default" which we no longer want to do. Thanks for flagging for a double check here.

PR #279, opened 2021-08-09
Add gray variant to `<Button />`

jmfury:
🎟️ [Asana Task](https://app.asana.com/0/1199897370927847/1200506511480647/f)
🔍 [Preview Link](https://react-components-git-{branch-slug}-hashicorp.vercel.app/components/button)

---

## Description

On Light (default): 
image

On dark (`background` set)
image

### PR Checklist 🚀

      1

### DIFF
--- a/packages/button/styles/variant-ghost.css
+++ b/packages/button/styles/variant-ghost.css
@@ -1,0 +1,63 @@
+.g-btn.variant-ghost {
+  color: var(--gray-3);
+  border-color: transparent;
+  background-color: transparent;
+
+  & .icon svg {
+    & [fill] {
+      fill: var(--gray-3);
+    }
+    & [stroke] {
+      stroke: var(--gray-3);
+    }
+  }
+
+  &.size-small,
+  &.size-medium {
+    padding: 0;
+  }
+
+  &:hover:not([disabled]) {
+    border-color: transparent;
+  }
+
+  &[disabled] {
+    &.background-light,
+    &.background-brand-secondary {
+      color: rgba(114, 114, 116, 0.3);
+      & .icon svg {
+        & [fill] {
+          fill: rgba(114, 114, 116, 0.3);
+        }
+        & [stroke] {
+          stroke: rgba(114, 114, 116, 0.3);
+        }
+      }
+    }
+    &.background-dark,
+    &.background-brand {
+      color: rgba(255, 255, 255, 0.3);
+      & .icon svg {
+        & [fill] {
+          fill: rgba(255, 255, 255, 0.3);
+        }
+        & [stroke] {
+          stroke: rgba(255, 255, 255, 0.3);
+        }
+      }
+    }
+  }
+
+  &.background-dark,
+  &.background-brand {
+    color: var(--white);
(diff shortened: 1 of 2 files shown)


packages/button/styles/variant-ghost.css
```diff
@@ -0,0 +1,63 @@
+.g-btn.variant-ghost {
+  color: var(--gray-3);
```
jmfury, line 2: Would have loved to just set a custom property here but given the rest of the SVG dependencies below, it felt like a little too fancy to introduce yet another styling pattern for this component

packages/button/styles/variant-ghost.css
```diff
@@ -0,0 +1,63 @@
...
+    &.background-brand {
+      color: rgba(255, 255, 255, 0.3);
+      & .icon svg {
+        & [fill] {
+          fill: rgba(255, 255, 255, 0.3);
+        }
+        & [stroke] {
+          stroke: rgba(255, 255, 255, 0.3);
+        }
+      }
```
jmfury, line 47: - [ ] Going to check on this

packages/button/styles/variant-ghost.css
alexcarpenter: Could the SVG dependencies below make use of `currentColor`? I could see that be the default in the index.css file and allow override if needed within each variant. Probably a bigger change than what you might want to do here, but it could reduce some of the duplication across the variants.

jmfury: Yeah that's a solid idea!

I would consider it a different effort since this component could also take advantage of css-modules and potentially some other upgrades / debt removal. 

css-modules alone feels like it might compound the amount of debt cleanup we'd need to do across sites as well.

packages/button/styles/variant-ghost.css
jmfury: I think it makes sense to keep this the same as `tertiary-neutral`

zchsh: +1, I think the contrast between the `disabled` state is good for both variants

alexcarpenter (commented):
Should we also add `ghost` as an option to the props docs? https://github.com/hashicorp/react-components/blob/435066c736c4887cd4417761030a652915993917/packages/button/props.js#L86

packages/button/styles/variant-ghost.css
zchsh: Just a +1 to all the above - between `currentColor` and potential use of custom properties since we dropped IE11 support, a refactor here could probably cut down on duplicated CSS quite a bit, and probably best as a separate PR & effort since it'll be a bigger change 🌊

jmfury:
> Should we also add `ghost` as an option to the props docs?
> 
> https://github.com/hashicorp/react-components/blob/435066c736c4887cd4417761030a652915993917/packages/button/props.js#L86

Great catch! Thank you!

zchsh (approved):
🎉 Looks great to me, thanks for picking this up 🙇

alexcarpenter (approved):
💥

PR #280, opened 2021-08-12
fix(content): fix incorrect body type class

zchsh:
🎟️ [Asana Task](https://app.asana.com/0/1100423001970639/1199185162145231/f)
🔍 [Preview Link](https://react-components-git-zsfix-content-blunder-hashicorp.vercel.app)

This PR fixes a facepalm-y issue in our `content` component, where `g-type-body-long` is composed, rather than the correct `g-type-long-body`.

### PR Checklist 🚀

### DIFF
--- a/packages/content/style.module.css
+++ b/packages/content/style.module.css
@@ -1,6 +1,6 @@
 .root {
   /* Base typography */
-  composes: g-type-body-long from global;
+  composes: g-type-long-body from global;
 
   /* Include all content base styles */
   composes: a from './styles/a.module.css';


BRKalow:
Haha oh man what an innocuous situation! I think your intuition is correct here in that it seems to make more sense to put it on the content container itself, not the individual elements. One, to enable more reliable cascading and general style within the content itself, like you said, and also to generally reduce unnecessary classes and processing.

Good catch! 👀

BRKalow (approved):

PR #283, opened 2021-08-13
feat(alert-banner)!: convert to CSS modules

zchsh:
🎟️ [Asana Task](https://app.asana.com/0/1100423001970639/1199185162145248/f)
🔍 [Preview Link](https://react-components-git-zsalert-banner-css-modules-hashicorp.vercel.app)

This PR converts `@hashicorp/react-alert-banner` to use CSS modules. It also refactors the component to use hooks, and migrates from Javascript to Typescript.

### Changes

- 💥✨ BREAKING CHANGE: Refactored to CSS modules.
  - Consumers will need to remove any `@hashicorp/react-alert-banner/style.css` imports.
- 🔨 Refactored from previous class component implementation to functional component with hooks
- 🐛 Minor tweak to styling, such that the `tag` and other text are always stacked.
  - This differs slight from the previous implementation, which would only show this behaviour if `tag` was longer than three characters.
  - This change was made to simplify the styling logic, with the rationale that the previous complexity was not worth the minor layout difference.
- 🐛 Removed the (attempted) collapse animation on close
  - This animation didn't seem to be working, since the state change to hide the banner resulted in a `display: none` style on the component before the animation could really start, much less complete.
  - We could add back this animation, but I didn't feel strongly that the additional complexity was worth it at this time.
  - If we do want to add back the collapse animation, we might be able to leverage [our Collapsible component](https://app.asana.com/0/1100423001970639/1157821186237305/f) (though it needs a hooks rewrite 😅 ).

### PR Checklist 🚀


packages/alert-banner/index.tsx
```diff
@@ -0,0 +1,98 @@
+import { useEffect, useState } from 'react'
+import cookie from 'js-cookie'
+import slugify from 'slugify'
+import classNames from 'classnames'
+import VisuallyHidden from '@reach/visually-hidden'
```
zchsh, line 5: I've added `VisuallyHidden` from ReachUI to replace one-off `.visuallyHidden` styles.

packages/alert-banner/style.module.css
```diff
@@ -1,5 +1,109 @@
...
+
+  &.isShown {
+    display: block;
+  }
+
+  /* Even if isShown is set,
+  should hide on mobile */
+  &.hideOnMobile {
+    @media (--small) {
+      display: none;
```
zchsh, line 15: `!important` no longer needed since styles are all together now 🎉

graphql.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module '*.graphql' {
```
zchsh, line 1: I encountered a TS error after the initial conversion of `alert-banner`:

CleanShot 2021-08-16 at 12 12 59@2x

Based on [this issue](https://github.com/apollographql/graphql-tag/issues/59), I _think_ this is the right solution for our use case. Wasn't sure whether it belonged somewhere else... Maybe https://github.com/hashicorp/nextjs-scripts/blob/main/packages/types/index.d.ts ?

Also noticed that https://github.com/hashicorp/hashicorp-www-next/blob/main/types/global.d.ts has something similar (possibly duplicative of `platform-types`?)

.github/pull_request_template.md
```diff
@@ -19,4 +19,3 @@ Items in this checklist may not may not apply to your PR, but please consider ea
 - [ ] Conduct reasonable cross browser testing for both compatibility and responsive behavior (We have a [Sauce Labs](https://app.saucelabs.com/) account for this, if you don't have access, just ask!).
 - [ ] Conduct reasonable accessibility review (use the [WAS](https://accessible.org/Web-Accessibility-Standards-WAS-2.pdf) as a guide or an [axe browser plugin](https://www.deque.com/axe/) until we establish more formal checks).
 - [ ] Identify (in the description above) and document (add Asana tasks on [this board](https://app.asana.com/0/1100423001970639/list)) any technical debt that you're aware of, but are not addressing as part of this PR.
-      1
```
zchsh, line 22: Very random but keeps bothering me 😅

zchsh:
@BRKalow For context here, main aim is CSS modules conversion. Tried my hand at a Typescript conversion as well, ran into a bit of an issue with `fragment.graphql` imports - I think the solution I have here works, but not sure if it's right. Would love any insight 🙇

graphql.d.ts
BRKalow: Yeah, I think this would be a good addition to our types package! https://github.com/hashicorp/nextjs-scripts/blob/main/packages/types/index.d.ts

packages/alert-banner/index.test.js
```diff
@@ -44,20 +52,22 @@ describe('<AlertBanner />', () => {
...
 
-    it('should NOT show the banner when the current date has surpassed the expiration date', () => {
+    it('should hide the banner after the expiration date', () => {
       const expirationDate = '2020-10-01T12:00:00-07:00'
       const { container } = render(
         <AlertBanner text="text" tag="tag" expirationDate={expirationDate} />
       )
-      expect(container.firstChild).not.toHaveClass('show')
+      const rootElem = container.firstChild
+      expect(rootElem.getAttribute('data-test-is-shown')).toBe('false')
```
BRKalow, line 70: We should still be able to assert on a class being present, we use https://www.npmjs.com/package/identity-obj-proxy in tests, so whatever key we're reading will be returned.

```jsx
<div className={s.isShown} /> // -> <div class="isShown" />
```

packages/alert-banner/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * On mount, hide the banner if it is expired or
+   * if a cookie indicates it was previously closed
+   */
+  useEffect(() => {
+    const hasBeenDismissed = cookie.get(dismissalCookieId)
+    const hasExpired = expirationDate && Date.now() > Date.parse(expirationDate)
+    const shouldBeShown = !hasBeenDismissed && !hasExpired
+    if (isShown !== shouldBeShown) setIsShown(shouldBeShown)
```
BRKalow, line 52: I think _technically_ `isShown` should be in the dependency array here. We can use a functional updater to avoid this, as well as drop the if and the need to reference `isShown` directly:

```tsx
setIsShown(current => current !== shouldBeShown ? shouldBeShown : current)
```

React will skip updates if the new value is the same:

> If your update function returns the exact same value as the current state, the subsequent rerender will be skipped completely.

https://reactjs.org/docs/hooks-reference.html#usestate

BRKalow (commented):
Your TS looks good, same with the CSS!

packages/alert-banner/index.test.js
zchsh: Oh nice, that's much better 💯 Will update 👍

BRKalow (approved):
👏

packages/alert-banner/index.test.js
zchsh: ✅ Updated

packages/alert-banner/index.tsx
zchsh: ✅ Updated

packages/alert-banner/index.test.js
```diff
@@ -1,36 +1,44 @@
...
-    expect(rootElem.tagName).toBe('DIV')
-    expect(rootElem).toHaveClass('g-alert-banner')
-  })
-
-  it('should render the right tag string', () => {
-    const { container } = render(<AlertBanner text="text" tag="tag" />)
-    expect(container.querySelector('.tag').textContent).toBe('tag')
-  })
+/**
+ * @TODO If we configured Jest to load CSS, we could convert some of these
```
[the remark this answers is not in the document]
zchsh: Update: removed the `data-test-is-shown` thanks to Bryce's tip, but I think this is still relevant 👍

graphql.d.ts
zchsh: 👍 - have opened https://github.com/hashicorp/nextjs-scripts/pull/198

PR #289, opened 2021-08-19
Combobox improvements

jmfury:
🎟️ [Asana Task]()

---

## Description
The proposed changes here allow for flexibility on the default  `<Combobox />` re:  `<input>`-related props and some sensible styling defaults for the list of options . More details about this PR outlined below, per use case.

#### Use case: `name` prop for certain forms
Without the `name` prop, the rendered `<input />` for the Combobox can't be serialized alongside other HTML input elements like you would expect. 

#### Use case: long lists of options could not be styled (to limit height / add scroll) from outside the default `<Combobox />`
We're using this for a long list of countries, and rather than forcing consumers to use `data-reach-*` selectors in their CSS, we provide a `--combobox-list-max-height` but also a sane default of `40vh` with scrolled overflow so we don't ever visually break the viewport with long lists

### PR Checklist 🚀

### DIFF
--- a/.changeset/sour-pears-exercise.md
+++ b/.changeset/sour-pears-exercise.md
@@ -1,0 +1,15 @@
+---
+'@hashicorp/react-combobox': major
+---
+
+Breaking Change: inputProps now handles all input related features to allow more flexibility
+
+### What changed?
+
+Previously `onInputChange` and `onInputBlur` were the only props consumers could use to customize the behavior of the `<input />` internally.
+
+`inputProps` now accepts all InputHTMLElement props for full customization of the input, and importantly allows for adding `name`, `id` or other standard attributes that might be required in certain form implementations.
+
+### How to update your code
+
+Please remove `onInputChange` and `onInputBlur` and use `inputProps` as an object to pass props to the internal `<input />` of the default `Combobox` component.
--- a/.changeset/sour-pears-exercise.md
+++ b/.changeset/sour-pears-exercise.md
@@ -28,6 +28,6 @@ Please remove `onInputChange` and `onInputBlur` and use `inputProps` as an objec
 +         onChange: handleInputChange,
 +         onBlur: validate,
 +         name: 'country',
-        }}
++       }}
         options={COUNTRY_OPTIONS}
 ```
--- a/.changeset/sour-pears-exercise.md
+++ b/.changeset/sour-pears-exercise.md
@@ -6,10 +6,28 @@ Breaking Change: inputProps now handles all input related features to allow more
 
 ### What changed?
 
-Previously `onInputChange` and `onInputBlur` were the only props consumers could use to customize the behavior of the `<input />` internally.
+Previously `onInputChange` and `onInputBlur` were the only props consumers could use to customize the behavior of the interal `<input />` of the `<Combobox />` (the default export of this package).
 
 `inputProps` now accepts all InputHTMLElement props for full customization of the input, and importantly allows for adding `name`, `id` or other standard attributes that might be required in certain form implementations.
+
+### Which components changed?
+
+- `<Combobox />` - the default export of this package - see above or below for me details
+- `<ComboboxInput />` - a named export of this package - now accepts `<input />`-specific props
 
 ### How to update your code
 
 Please remove `onInputChange` and `onInputBlur` and use `inputProps` as an object to pass props to the internal `<input />` of the default `Combobox` component.
+
+```diff
+      <Combobox
+        label="Select your country"
+-       onInputChange={handleInputChange}
+-       onInputBlur={validate}
++       inputProps={{
++         onChange: handleInputChange,
++         onBlur: validate,
++         name: 'country',
+        }}
+        options={COUNTRY_OPTIONS}
+```
--- a/packages/combobox/index.tsx
+++ b/packages/combobox/index.tsx
@@ -19,9 +19,11 @@ export interface ComboboxProps {
   renderOption?: (option: ComboboxOptionValue) => ReactNode
   options: ComboboxOptionValue[]
   openOnFocus?: boolean
-  inputProps: ComboboxInputProps
+  inputProps: Optional<ComboboxInputProps, 'onChange'>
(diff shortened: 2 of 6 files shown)


packages/combobox/primitives.tsx
```diff
@@ -49,7 +49,9 @@ export function ComboboxPopover({
   )
 }
 
-export interface ComboboxInputProps extends ReachComboboxInputProps {
+type InputProps = InputHTMLAttributes<HTMLInputElement> &
+  ReachComboboxInputProps
+export interface ComboboxInputProps extends InputProps {
```
jmfury, line 54: @BRKalow  - Wanted to get your take on the types here. 

I think I need to potentially split this once more (maybe with `Omit`?) so that I can make [`onChange` optional here at the top level (the default export, `<Combobox />`)](https://github.com/hashicorp/react-components/pull/289/files#diff-2d04980ac40707a6752abd67b8b92f569f0582630476c1c3d89881286419a52fR22) since there's a chance it's not necessary if folks only require `onSelect` for their use case

BRKalow: From: https://stackoverflow.com/questions/43159887/make-a-single-property-optional-in-typescript

```ts
type Optional<T, K extends keyof T> = Pick<Partial<T>, K> & Omit<T, K>;

// ...

Optional<ComboboxInputProps, 'onChange'>
```

Or just make it always optional and provide an empty fn as a default:
```
onChange: () => {}
```

packages/combobox/props.js
```diff
@@ -14,13 +14,23 @@ module.exports = {
...
-  onInputChange: {
-    type: 'function',
-    description: 'A change handler for the input',
-  },
-  onInputBlur: {
-    type: 'function',
-    description: 'An `onBlur` handler for the input',
+  inputProps: {
+    type: 'object',
+
```
jmfury, line 19:
```suggestion
```

.changeset/sour-pears-exercise.md
```diff
@@ -0,0 +1,15 @@
...
+
+### What changed?
+
+Previously `onInputChange` and `onInputBlur` were the only props consumers could use to customize the behavior of the `<input />` internally.
+
+`inputProps` now accepts all InputHTMLElement props for full customization of the input, and importantly allows for adding `name`, `id` or other standard attributes that might be required in certain form implementations.
+
+### How to update your code
+
+Please remove `onInputChange` and `onInputBlur` and use `inputProps` as an object to pass props to the internal `<input />` of the default `Combobox` component.
```
kendallstrautman, line 15: A diff example here could be helpful!

packages/combobox/index.tsx
```diff
@@ -18,11 +19,12 @@ export interface ComboboxProps {
   renderOption?: (option: ComboboxOptionValue) => ReactNode
   options: ComboboxOptionValue[]
   openOnFocus?: boolean
-  onInputChange?: (e: React.ChangeEvent<HTMLInputElement>) => void
-  onInputBlur?: (e: React.FocusEvent<HTMLInputElement>) => void
+  inputProps: Optional<ComboboxInputProps, 'onChange'>
```
kendallstrautman, line 22: Sorry if I just don't have context but am curious why not make it optional on the type itself? Is there another instance where it's required?

packages/combobox/index.tsx
```diff
@@ -18,11 +19,12 @@ export interface ComboboxProps {
...
   invalidInputValue?: boolean
 }
 
+type Optional<T, K extends keyof T> = Pick<Partial<T>, K> & Omit<T, K>
```
kendallstrautman, line 26: Handy!

packages/combobox/props.js
```diff
@@ -14,13 +14,22 @@ module.exports = {
...
-    type: 'function',
-    description: 'A change handler for the input',
-  },
-  onInputBlur: {
-    type: 'function',
-    description: 'An `onBlur` handler for the input',
+  inputProps: {
+    type: 'object',
+    description:
+      "Props passed to the combobox input, props listed below are not an exhaustive list, conforms to React's InputHTMLAttributes",
```
kendallstrautman, line 20: Could we link to an exhaustive list example? If it exists.

.changeset/sour-pears-exercise.md
```diff
@@ -0,0 +1,15 @@
+---
+'@hashicorp/react-combobox': major
+---
+
+Breaking Change: inputProps now handles all input related features to allow more flexibility
+
+### What changed?
+
+Previously `onInputChange` and `onInputBlur` were the only props consumers could use to customize the behavior of the `<input />` internally.
```
kendallstrautman, line 9: Since this package exports so many different components, could you clarify which ones are affected?

kendallstrautman (approved):
Seems like reasonable interface updates! 👏

Check if you need to update any docs examples! I didn't see any diffs in that file.

packages/combobox/index.tsx
jmfury: That is a really great point! -- I'll take a look, I think it probably can be optional at base

packages/combobox/props.js
jmfury: Yep, I thought about it, maybe https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts -- I hesitate to put a line number permalink since it could change

packages/combobox/style.module.css
```diff
@@ -19,6 +19,8 @@
       box-shadow: var(--select-shadow);
       background: var(--white);
       border-radius: 2px;
+      max-height: var(--combobox-list-max-height, 40vh);
+      overflow: scroll;
```
alexcarpenter, line 23: @jmfury Can we use `overflow-y: auto` here? That way the scroll bar never shows horizontally and only vertically when needed.

jmfury: @alexcarpenter Great call - silly oversight on my part. Can you to throw up a PR for a patch? -- if not I can, but won't get to it til Monday -- moving onto some other stuff atm!

PR #292, opened 2021-08-20
[downloads-page] Add enterprise package managers for linux

jescalan:
## Description

@kpenfound was very helpful in providing information about supported package managers for ent binaries, this adds them in to ent mode!

### PR Checklist 🚀

### DIFF
--- a/next-env.d.ts
+++ b/next-env.d.ts
@@ -1,3 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/types/global" />
 /// <reference types="next/image-types/global" />
+
+// NOTE: This file should not be edited
+// see https://nextjs.org/docs/basic-features/typescript for more information.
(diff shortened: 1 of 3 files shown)


next-env.d.ts
```diff
@@ -1,3 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/types/global" />
 /// <reference types="next/image-types/global" />
+
+// NOTE: This file should not be edited
+// see https://nextjs.org/docs/basic-features/typescript for more information.
```
BRKalow, line 6: Did you add this or, was it also generated?

BRKalow (approved):
LGTM! Needs a changeset file

next-env.d.ts
jescalan: it was generated - i was also surprised by this

BRKalow: Probably helpful regardless 😄 I was surprised when it started getting unconditionally overwritten with Next 11.

PR #309, opened 2021-08-28
feat(logo-grid): convert to CSS modules

zchsh:
🎟️ [Asana Task](https://app.asana.com/0/1100423001970639/1199185162145294/f)
🔍 [Preview Link](https://react-components-git-zslogo-grid-css-modules-hashicorp.vercel.app/components/logogrid)

---

This PR converts `@hashicorp/react-logo-grid` to CSS modules. One significant change to enable this shift was the removal of [Tippy.js](https://github.com/atomiks/tippyjs-react), which has been replaced by a [ReachUI Dialog](https://reach.tech/dialog) element.

### PR Checklist 🚀

> - [ ] ~Add or update tests as appropriate.~
    - Note: test file had only `.todo`s. For now I've left these as-is, have **not** added new tests yet.
    - Rationale for not fixing the test `todo`s yet...
        1. component is mostly presentation, and have done a good amount of manual QA
        2. I think we would prefer to keep momentum on CSS modules conversions rather than rethink our testing strategy for this component.
    - ☝️ That being said, loosely held approach here. Very open to sinking some extra time in to get some good tests written here. (I mean I am open to it either way, just a question of whether we do it now, in this PR, or later, as a follow-up)


packages/logo-grid/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+          return <CompanyLogo company={company} color={color} size={size} />
+        }
+        return (
+          <li
+            key={company.name}
+            className={classNames(s.listItem, s[size], {
+              [s.removeBorders]: removeBorders,
+            })}
+          >
+            {hasLink ? (
```
zchsh, line 82: Not super proud of this nested ternary, but it felt like a decent way to avoid prop-drilling. Very open to other ideas on how to clean this up 🙇

packages/logo-grid/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+            linkType="outbound"
+            external={true}
+          />
+        )}
+      </DialogTooltip>
+    </>
+  )
+}
+
+function CompanyLogo({ company, color, size }) {
```
zchsh, line 148: Split this component out, similar to in the previous implementation.

packages/logo-grid/partials/dialog-tooltip/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+  triggerRect: DOMRect
+  /** */
+  arrowSize?: number
+  /** Minimum distance in pixels that the dialog should be from the viewport edge.  */
+  collisionBuffer?: number
+  /** Color scheme appearance of the component. Works best in contexts with a matching theme. */
+  theme?: 'light' | 'dark'
+}
+
+function DialogTooltip({
```
zchsh, line 28: Similar to the Reach-based `tooltip` created for `tabs`, I split this out in the hopes that it might someday see re-use. I did not try to consolidate with the `tabs` tooltip - in this case, we have interactive content in our tooltip, so we need to use a more Dialog-like component.

packages/logo-grid/partials/dialog-tooltip/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+  )
+
+  /* Odd fix for possible issue with useRect, where
+  tooltipRect did not seem to be set as expected.
+  I think this works because it forces a re-render
+  when "shown" changes, which is exactly when we need
+  to look at the tooltipRef again and re-measure.
+  Without this, tooltipRect seems to always be null.
+  Not actually sure why this works. Need to investigate. */
+  useEffect(setNow, [shown])
```
zchsh, line 58: I need help understanding this 😅 It _feels_  like `useRect` should just work without this, and update properly with the dialog's `DOMRect` when it is opened (and therefore rendered). I'm not sure why  it  doesn't. Possibly related to `Portal` use, as  I did test the same thing without `Portal` and it worked fine. cc @BRKalow @jescalan @thiskevinwang in case y'all have any insight 🙇

packages/logo-grid/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+}
+
+function LogoGrid({
+  data,
+  color = 'color',
+  details,
+  integrationLink,
+  removeBorders,
+  size = 'small',
+  className,
```
zchsh, line 63: I've added a `className` prop. Rationale here is  that [a quick SourceGraph search](https://sourcegraph.hashi-mktg.com/search?q=g-logo-grid&patternType=literal) can easily determine whether the old global className was targeted in "overrides" - if so, I immediately add a `className` prop so that we can, for now, retain these overrides rather than having their replacement block rollout of these components.

In cases where there are no overrides, ie that SourceGraph search comes up empty, I could still see a case for adding a `className` prop to enable "safe overrides" (as detailed in the Notion doc on CSS). So if I ever find a component that has 0 overrides in the wild, may still be worth adding this 👍

packages/logo-grid/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+  logo?: CompanyLogo
+  /** Monochrome black logo image. */
+  monochromeLogo?: CompanyLogo
+  /** Monochrome white logo image. */
+  whiteLogo?: CompanyLogo
+}
+
+interface LogoGridProps {
+  /** Array of company objects to render as grid items. */
+  data: Array<Company>
```
BRKalow, line 41: nit: The more conventional way to define an array type is:
```suggestion
  data: Company[]
```

packages/logo-grid/partials/dialog-tooltip/index.tsx
```diff
@@ -0,0 +1,195 @@
+import React, { useState, useRef, useEffect } from 'react'
+import { DialogOverlay, DialogContent } from '@reach/dialog'
```
BRKalow, line 2: I don't mean to throw a wrench in your progress here, but do you think https://www.npmjs.com/package/@reach/popover would be more appropriate for this use case?

BRKalow (commented):
Fantastic documentation, as usual! Left a note about exploring `@reach/popover`, which I think might be more appropriate for this pattern.

.changeset/smart-countries-fly.md
```diff
@@ -0,0 +1,13 @@
+---
+'@hashicorp/react-logo-grid': major
+---
+
+- 💥✨ BREAKING CHANGE: Converts to CSS modules.
+  - Consumers will need to remove any `@hashicorp/react-logo-grid` imports.
```
jescalan, line 6:
```suggestion
  - Consumers will need to remove any `@hashicorp/react-logo-grid` css imports.
```

.changeset/smart-countries-fly.md
```diff
@@ -0,0 +1,13 @@
...
+  - No longer renders a `g-logo-grid` className. Does however accept a `className` prop, so that we can continue to meet override use cases.
+- 🔨 Replaces Tippy with ReachUI Dialog
+  - This changes the behavior of the dialog somewhat, but not in a breaking way.
+- 🔨 Accessibility improvements
+  - grid items with dialog tooltips now render as focus-able `button` elements
+  - tooltip a11y is now improved, with features such as focus lock and auto-focusing of the close button when the dialog is opened, thanks to ReachUI
+- 🔨 Converts to Typescript
```
jescalan, line 13: 🎉

packages/logo-grid/index.tsx
jescalan: Only idea I have would be to make it a separate function component, then you could use normal if/else statements. But this doesn't really feel that bad honestly.

packages/logo-grid/partials/dialog-tooltip/index.tsx
jescalan: This one doesn't appear on their website 🤔

packages/logo-grid/partials/dialog-tooltip/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+  return (
+    <>
+      <DialogOverlay
+        className={classNames(s.dialogOverlay, s[theme])}
+        isOpen={shown}
+        onDismiss={() => setIsShown(false)}
+        /* We don't want scroll lock, as this is more of a tooltip than a dialog.
+        Due to our tooltip-like positioning, when scroll lock is enabled, the
+        tooltip can become locked in an offscreen position. */
+        dangerouslyBypassScrollLock={true}
```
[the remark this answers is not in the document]
jescalan: agreed and this seems ok to me in this instance

packages/logo-grid/partials/dialog-tooltip/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+          {
+            '--left': arrowLeft,
+            '--top': arrowTop,
+            '--arrow-size': arrowSize + 'px',
+          } as React.CSSProperties
+        }
+      />
+    </Portal>
+  )
+}
```
jescalan, line 193: My only thoughts here are that this is a lot of complicated custom code - I'm not sure that its avoidable, it almost feels like we removed tippy only to effectively re-implement it manually

jescalan (commented):
I'm curious about the popover element bryce mentioned - this is GOOD, as expected, but just feels like a lot of manual implementation

packages/logo-grid/partials/dialog-tooltip/index.tsx
BRKalow: Yeah, not sure why that is exactly. I do think it's the right component for this use though. It looks like they use it as part of their combobox component.

zchsh: Ah nice - yeah support for the `position` prop could definitely help clean things up a little and make this feel a little less like a re-implementation of Tippy. Hopefully it'll resolve the annoying `useRect` issue I was running into as well. Will give it a shot 👍

zchsh: @BRKalow I gave `@reach/popover` a shot - did up needing some workarounds, but now have both `DialogTooltip` and `PopoverTooltip` with identical APIs which we can choose from. A quick summary:

- `DialogTooltip` uses `@reach/dialog`, which is [well-documented](https://reach.tech/dialog), and aligns somewhat with reach.tech's recommendation to use `dialog` for "interactive tooltips". By using `dialog`, we get focus behaviour, keyboard and on-click-outside functionality to dismiss the dialog by default. But, the implementation required significant workarounds, including custom positioning, and a weird effect to force re-rendering whenever the dialog is opened or closed.

- `PopoverTooltip` uses `@reach/popover`. It seems to have no official documentation - unclear why this is. By using `popover`, we get focus behaviour and custom positioning by default. But, the implementation required the use of `onClickOutside` and `onFocusOutside` to allow the popover to be dismissed as expected when a visitor clicks or focuses away.

Leaning towards using `@reach/popover` as you recommended, wanted to run this by just in case. If that sounds good, will just commit the deletion of the `dialog`-based implementation and this should be good to go 👍 🙇

BRKalow: Thanks for spiking this out. The popover implementation seems potentially cleaner...what do you think? It seems better to have to handle the click/focus outside than the custom positioning. I'll let you make the final call here though!

zchsh:
@BRKalow @jescalan Last component conversion to merge 🎉 Have switched in `@reach/popover` here - details in [this comment](https://github.com/hashicorp/react-components/pull/309#discussion_r704686783). Open to thoughts here. I'll interpret PR approval as a signal we should go with the `popover` implementation and drop the `dialog` one 👍

BRKalow (approved):
Happy with Dialog or Popover, though the popover implementation seems a little cleaner to me!

packages/logo-grid/partials/dialog-tooltip/index.tsx
zchsh: +1 I think `popover` is a bit cleaner - will go with that unless @jescalan has any strong objections 👍

jescalan (approved):
I'm good with whichever option you both prefer, which sounds like popover. A little uncomfortable with the amount of custom code involved with both, but it's probably better than tippy, right?

zchsh:
@jescalan Yep, I think it's still more lightweight than Tippy. The "tooltip arrow" part feels like a decent amount of custom code here... I expect if we continue to see tooltip patterns in design (which seems very likely) we can refine things further, and hopefully make the custom parts feel a little more contained 👍

PR #312, opened 2021-08-28
refactor(text-input)!: convert to CSS modules

zchsh:
🎟️ [Asana Task](https://app.asana.com/0/1100423001970639/1199185162145310/f)
🔍 [Preview Link](https://react-components-git-zstext-input-css-modules-hashicorp.vercel.app/components/textinput)

---

This PR updates `@hashicorp/react-text-input` to use CSS modules.

### PR Checklist 🚀


packages/text-input/index.js
```diff
@@ -1,42 +1,48 @@
+import classNames from 'classnames'
 import { v1 as uuidv1 } from 'uuid'
+import s from './style.module.css'
 
-function TextInput(props) {
-  //  `field` and `form` props come from Formik
-  //  https://jaredpalmer.com/formik/docs/api/field
-  const { field, form, type, label, placeholder, theme } = props
```
zchsh, line 6: Moved prop de-structuring into the function argument area, to align with our approach in other components.

packages/text-input/index.js
```diff
@@ -1,42 +1,48 @@
...
+  label,
+  placeholder,
+  theme = {
+    background: 'light',
+  },
+}) {
+  //  Note: `field` and `form` props typically come from Formik
+  //  ref: https://jaredpalmer.com/formik/docs/api/field
+  //  However, they can also be provided in non-Formik use cases
+  //  and this component will still function as expected.
```
zchsh, line 19: Added this note to try to help clarify that it's totally fine to use this component without Formik.

packages/text-input/index.test.js
```diff
@@ -6,16 +6,19 @@ import { getTestValues } from 'swingset/testing'
...
     const rootElem = container.firstChild
     expect(rootElem.tagName).toBe('DIV')
-    expect(rootElem).toHaveClass('g-text-input')
+    expect(rootElem).toHaveClass(className)
   })
 
   it('should hide the root element when `type` is set to `hidden`', () => {
     const { container } = render(<TextInput {...defaultProps} type="hidden" />)
-    expect(container.firstChild).not.toBeVisible()
+    expect(container.firstChild).toHaveClass('hidden')
```
zchsh, line 21: Wasn't able to maintain this test with `toBeVisible`, possibly due to differences in how we load normal CSS vs CSS modules? While not perfect, testing for the `hidden` class feels like a decent alternative.

packages/text-input/package-lock.json
```diff
@@ -0,0 +1,126 @@
+{
+  "name": "@hashicorp/react-text-input",
```
zchsh, line 2: Wanted to check on this - noticed that in some PRs, it seems like I'm adding back a `package-lock.json` into the component's folder. Is this the correct approach? Or should these component-level `package-lock` files be deleted?

Did some quick searching and found [this suggestion that maybe I should've installed from the project root](https://stackoverflow.com/questions/65237564/npm-7-workspaces-how-to-install-new-package-in-workspace). Assuming that's the right approach, then it seems I should make sure my PRs are **not** adding component-level `package-lock.json` files. cc @BRKalow @jescalan @thiskevinwang to double-check on this 🙇

pages/global.css
```diff
@@ -14,7 +14,6 @@
 @import '../packages/content-cta/style.css';
 @import '../packages/consent-manager/style.css';
 @import '../packages/toggle/style.css';
-@import '../packages/text-input/style.css';
```
zchsh, line 17: 🔪 🎉

packages/text-input/index.js
```diff
@@ -1,42 +1,48 @@
...
-        <label htmlFor={inputId}>
-          <span className="g-type-label">{label}</span>
+        <label htmlFor={inputId} className={s.label}>
+          {label}
         </label>
       )}
       <input
         id={inputId}
         type={type}
-        data-has-error={(!!error).toString()}
```
BRKalow, line 25: Was this data attr just used for styling previously?

packages/text-input/style.module.css
```diff
@@ -0,0 +1,61 @@
...
+
+  &.hasError {
+    border-color: var(--danger);
+  }
+}
+
+.error {
+  composes: g-type-body-small from global;
+  color: var(--danger);
+  line-height: 1.4;
```
BRKalow, line 59: It seems slightly arbitrary that the error `line-height` is a tiny bit different than the input's (`1.4` for error vs. `1.5em` for the input). 🤔 I wonder if we could align these.

packages/text-input/index.js
zchsh: Yep, exactly. Also this question is a great example of why we decided we should _not_ use `data-` attributes for styling 😅

packages/text-input/style.module.css
zchsh: Great callout - it does feel a little arbitrary, but I think this might be more of a design concern. Assuming the viewer hasn't scaled their font-size, the `1.4` value also gives us a round `21px` rather than `22.5px`. Quick preview:

CleanShot 2021-09-07 at 17 24 43@2x

While I definitely don't feel there's a big difference either way, I also don't feel like aligning the numbers gives us a very tangible benefit, and worry that we might be making a (very slight) compromise to design polish.

jescalan (approved):

packages/text-input/package-lock.json
zchsh: Going to handle this in a follow-up PR once all CSS modules conversion PRs are merged (added a new task to the project board for this). I suspect the right move will be to clean up any `package-lock.json` files within component package directories, but will confirm in that upcoming PR 👍

BRKalow (approved):

PR #313, opened 2021-08-28
feat(text-split)!: convert to CSS modules

zchsh:
🎟️ [Asana Task](https://app.asana.com/0/1100423001970639/1199185162145312/f)
🔍 [Preview Link](https://react-components-git-zstext-split-css-modules-hashicorp.vercel.app/components/textsplit)

---

This PR updates `@hashicorp/react-text-split` to use CSS modules.

### PR Checklist 🚀


packages/text-split/index.js
```diff
@@ -19,15 +21,12 @@ function TextSplit({
   const hasReactContent = content && !hasStringContent
 
   return (
-    <div className={`g-text-split background-${theme}`}>
-      <div className={`g-grid-container text-at-${textSide}`}>
-        <div className="children">{children}</div>
-        <div className="text">
+    <div className={classNames(s.root, s[`background-${theme}`])}>
```
BRKalow, line 24: Should we also support a top-level `className` prop here?

pages/global.css
```diff
@@ -20,7 +20,6 @@
 @import '../packages/vertical-text-block-list/style.css';
 @import '../packages/tabs/style.css';
 @import '../packages/tabbed-accordion/style.css';
-@import '../packages/text-split/styles/index.css';
```
BRKalow, line 23: :feelsgood:

packages/text-split-with-code/index.test.js
```diff
@@ -3,6 +3,7 @@ import TextSplitWithCode from './'
 
 const baseProps = {
   textSplit: {
+    className: 'g-text-split',
```
zchsh, line 6: Have updated `text-split-with-*` tests to retain their existing behaviour, by using the newly added `className` prop on `text-split`.

zchsh:
Just noticed - the `changeset` in this PR doesn't seem to be registering 🤔 Will try to regenerate the changeset file...

packages/text-split/index.test.js
```diff
@@ -21,11 +21,14 @@ const propsBase = {
...
+  it('should add a provided className to the root element', () => {
+    const className = 'my-nice-text-split'
+    const { container } = render(
+      <TextSplit {...propsBase} className={className} />
+    )
     const rootElem = container.firstChild
     expect(rootElem.tagName).toBe('DIV')
-    expect(rootElem).toHaveClass('g-text-split')
+    expect(rootElem).toHaveClass(className)
   })
```
BRKalow, line 32: Just a meta observation across all of these PRs, I wonder if a test helper for this would be worthwhile:
```js
function assertAcceptsRootClassName(Component, className, props) {
  const { container } = render(<Component {...props} />)
  expect(container.firstChild).toHaveClass(className)
}

it ('should add a provided className to the root element', () => assertAcceptsRootClassName(TextSplit, className, props))
```

BRKalow (approved):

packages/text-split/index.test.js
zchsh: Great point. Personally, somewhat on the fence about this - we used to have really heavily abstracted test code, and sometimes it felt like it made it much more difficult to understand and contribute to tests (unless you were already familiar with the abstractions). This feels super manageable though, so I might just be biased / haunted by that past experience 😅

PR #321, opened 2021-09-01
feat(min-100-layout): add Min100Layout component

zchsh:
🎟️ [Asana Task](https://app.asana.com/0/1200392284811676/1200892636887647/f)
🔍 [Preview Link](https://react-components-git-zsmin-100-layout-component-hashicorp.vercel.app/components/min100layout)

---

This PR adds a `@hashicorp/react-min-100-layout` component. This layout component is designed for use in `_app.tsx` or in top-level layouts, and ensures pages with footers will occupy the full viewport height even if their main page content is not as tall.

One example use case, where the layout would likely look much more natural if the main content occupied enough space to push the footer lower on the page:
https://boundary-git-fork-zchsh-zstest-packer-openapi-docs-hashicorp.vercel.app/api-docs-by-service

Learn already implements a similar approach with one-off global styles:
https://github.com/hashicorp/learn/blob/efe39da15b5b75fbc2aa2428b6f5d45a1a89fa71/pages/style.css#L266

### PR Checklist 🚀

### DIFF
--- a/packages/min-100-layout/index.test.js
+++ b/packages/min-100-layout/index.test.js
@@ -1,0 +1,25 @@
+import { render, screen } from '@testing-library/react'
+import Min100Layout from '.'
+
+describe('<Min100Layout />', () => {
+  it('should pass a provided className to the root element', () => {
+    const className = 'my-special-layout-class'
+    const { container } = render(
+      <Min100Layout className={className} footer={<p>footer</p>}>
+        <p>body</p>
+      </Min100Layout>
+    )
+    const rootElem = container.firstChild
+    expect(rootElem).toHaveClass(className)
+  })
+
+  it('should render the provided footer and body content', () => {
+    render(
+      <Min100Layout footer={<p>footer</p>}>
+        <p>body</p>
+      </Min100Layout>
+    )
+    expect(screen.getByText('body')).toBeInTheDocument()
+    expect(screen.getByText('footer')).toBeInTheDocument()
+  })
+})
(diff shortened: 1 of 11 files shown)


packages/min-100-layout/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+  footer: React.ReactNode
+}
+
+function Min100Layout({
+  children,
+  footer,
+}: Min100LayoutProps): React.ReactElement {
+  return (
+    <div className={s.root}>
+      <div>{children}</div>
```
zchsh, line 21: Wrapped `children` here because rendering directly into a `flex` container felt like it might be a bit unexpected.

packages/min-100-layout/style.module.css
```diff
@@ -0,0 +1,10 @@
+.root {
```
zchsh, line 1: These styles are based on a [similar solution implemented for Learn](https://github.com/hashicorp/learn/blob/2ded2a4d21e4649c9bdcf6afc0824fd23cfcbba4/pages/style.css#L266).

packages/min-100-layout/index.test.js
```diff
@@ -0,0 +1,25 @@
+import { render, screen } from '@testing-library/react'
+import Min100Layout from '.'
+
+describe('<Min100Layout />', () => {
+  it('should pass a provided className to the root element', () => {
```
zchsh, line 5: This is a pretty minimal render-some-CSS component, so these are pretty minimal tests.

jescalan (approved):

zchsh:
Proved this out in https://github.com/hashicorp/learn/pull/3861, so going to go ahead and merge 👍

PR #322, opened 2021-09-01
`<InlineSvg / >` Allow span element wrapper

alexcarpenter:
🎟️ [Asana Task](https://app.asana.com/0/1200392284811676/1200899791949968)

---

## Description

### PR Checklist 🚀

### DIFF
--- a/packages/inline-svg/index.js
+++ b/packages/inline-svg/index.js
@@ -1,3 +1,9 @@
-export default function InlineSvg({ src, ...props }) {
-  return <div dangerouslySetInnerHTML={{ __html: src }} {...props}></div>
+const supportedElements = {
+  div: 'div',
+  span: 'span',
+}
+
+export default function InlineSvg({ wrapper, src, ...props }) {
+  const Component = supportedElements[wrapper] || supportedElements['div']
+  return <Component dangerouslySetInnerHTML={{ __html: src }} {...props} />
 }


packages/inline-svg/index.test.js
```diff
@@ -47,4 +47,17 @@ describe('<InlineSvg />', () => {
     expect(rootElem.localName).toBe('div')
     expect(rootElem.childNodes.length).toBe(0)
   })
+
+  fit('should render custom wrapper element', () => {
```
jmfury, line 51:
```suggestion
  it('should render custom wrapper element', () => {
```
Right? 😅

packages/inline-svg/index.js
```diff
@@ -1,3 +1,9 @@
-export default function InlineSvg({ src, ...props }) {
-  return <div dangerouslySetInnerHTML={{ __html: src }} {...props}></div>
+const supportedElements = {
+  div: 'div',
+  span: 'span',
+}
+
+export default function InlineSvg({ wrapper, src, ...props }) {
+  const Component = supportedElements[wrapper] || supportedElements['div']
```
jescalan, line 7: what happens if someone doesn't enter in `span` or `div` here?

packages/inline-svg/index.js
```diff
@@ -1,3 +1,9 @@
-export default function InlineSvg({ src, ...props }) {
-  return <div dangerouslySetInnerHTML={{ __html: src }} {...props}></div>
+const supportedElements = {
+  div: 'div',
+  span: 'span',
+}
+
+export default function InlineSvg({ wrapper, src, ...props }) {
```
jmfury, line 6:
```suggestion
export default function InlineSvg({ wrapper = 'div', src, ...props }) {
```

Have you tried this w/ `yalc` yet? I think TS will complain about `wrapper` if you try not to define it in the implementation w/o a default or explicit optional in an interface

jescalan (commented):
I like this direction but feel like we maybe need to take a harder stance on what is supported. This implementation feels like it's meant to be flexible in case we need to support other wrapper elements in the future, but as a result makes the API more explicit and more prone to error.

I would just offer up for discussion - is it feasible to make the prop `spanWrapper` or something more explicit like that, and still have it be ok? I am trying to imagine a case where we'd need to have the wrapping element be something else and I can't really come up with one.

packages/inline-svg/index.js
alexcarpenter: Currently only `div` and `span` are supported, if an element other than the `supportedElements` is provided, we fallback to `div`. Agreed we can make a more explicit contract using `spanWrapper` as a boolean.

packages/inline-svg/index.js
alexcarpenter: Not familiar with `yalc`. Not seeing it in scripts, is this a tool outside of the project I should be using?

PR #323, opened 2021-09-01
Adds optional lang prop to `<HashiHead />`

alexcarpenter:
🎟️ [Asana Task](https://app.asana.com/0/1200641486648086/1200899791949972)

---

## Description

This update allows us to set an optional lang attribute on the `<html />` element via the `<HashiHead />` component.

### PR Checklist 🚀

### DIFF
--- a/packages/head/index.tsx
+++ b/packages/head/index.tsx
@@ -3,6 +3,7 @@ import Head from 'next/head'
 export default function HashiHead(props: HashiHeadProps): React.ReactElement {
   return (
     <Head>
+      {whenString(props.lang, <html lang={props.lang} />)}
       {whenString(props.title, <title>{props.title}</title>)}
       <meta httpEquiv="x-ua-compatible" content="ie=edge" />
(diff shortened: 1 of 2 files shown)


packages/head/index.tsx
```diff
@@ -3,6 +3,7 @@ import Head from 'next/head'
 export default function HashiHead(props: HashiHeadProps): React.ReactElement {
   return (
     <Head>
+      {whenString(props.lang, <html lang={props.lang} />)}
```
BRKalow, line 6: Just a quick check-in, does this work? the `html` tag would wrap `head`, so I would be surprised if this could mutate that tag.

alexcarpenter: @BRKalow It does work, but I have been sitting on whether or not this would be the right solution vs getting the `i18n` next.config setting to work with cloud. Was getting an error on deployments with `next export`.

alexcarpenter:
Closing out in favor of finding an alternative solution.

PR #328, opened 2021-09-08
Add package manager config

jescalan:
🎟️ [Asana Task]()

---

## Description

Many moons ago, I added a feature to swingset that allowed for package managers to be added next to component names, and then forgot to actually implement it here. Well, here it is!

### PR Checklist 🚀

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28249,7 +28249,7 @@
     },
     "packages/accordion": {
       "name": "@hashicorp/react-accordion",
-      "version": "2.0.3",
+      "version": "3.0.0",
       "license": "MPL-2.0",
       "dependencies": {
         "@hashicorp/react-inline-svg": "^6.0.2",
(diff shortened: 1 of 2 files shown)


BRKalow:
Wow I remember this! Good catch. I think we might need to change the slug generation to produce snake case...multi word components are producing broken links

jescalan:
@BRKalow great catch thank you! guess i should have tested more than just `Button` 🤦 - should be fixed now

BRKalow (approved):
Looks good!

I wonder if exposing the `package.json` data and embedding the filepath of the component into the swingset data somehow would be a more robust solution here. I don't think we need to do that now, but something to noodle on. I think we do something similar for the "Where it's used" section.

```
customMeta({ data }) {
  return {
    github: `https://github.com/hashicorp/react-components/tree/main/${data.filepath}`,
    npm: `https://npmjs.com/package/${packageJson.name}`
  }
}
```

PR #336, opened 2021-09-14
DocsSidenav - add headline option

jescalan:
🔍 [Preview Link](https://react-components-e8inu5aoi-hashicorp.vercel.app/components/docssidenav)

---

## Description

We need the capability to add small headlines to the sidenav for Terraform, this is a nice simple implementation of it.

### PR Checklist 🚀

### DIFF
--- a/.changeset/small-dingos-change.md
+++ b/.changeset/small-dingos-change.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-docs-sidenav': minor
+---
+
+Add "headline" option which can be used to inject a headline at any point in the navigation
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,9 +7,7 @@
     "": {
       "name": "@hashicorp/react-components",
       "version": "0.0.0",
-      "workspaces": [
-        "./packages/*"
-      ],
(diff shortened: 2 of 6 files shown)


BRKalow (changes requested):
Can we also update the `types.ts` file with the new heading node type?

https://github.com/hashicorp/react-components/blob/main/packages/docs-sidenav/types.ts

EnMod (changes requested):
Code looks good, nice and simple :+1: 

It seems that the search filter is now broken though, entering in any value throws this error:

`TypeError: Cannot read property 'toLowerCase' of undefined`

The `filterContent` util function seems to be the culprit, specifically this check at [line 11](https://github.com/hashicorp/react-components/blob/main/packages/docs-sidenav/utils/filter-content.js#L11):

```javascript
const isTitleMatch = fuzzysearch(
  searchValue.toLowerCase(),
  item.title.toLowerCase()
)
```

I'm thinking now it needs to account for the possibility of headings?

BRKalow (approved):
👍  nice!

EnMod (approved):
Cool, LGTM!

PR #337, opened 2021-09-14
chore: DocsPage to TypeScript

thiskevinwang:
🎟️ [Asana Task](https://app.asana.com/0/1100423001970639/1200982361585779/f)
🔍 [Preview Link](https://react-components-hpkzl51d7-hashicorp.vercel.app/components/docspage)

---

## Description

This converts `DocsPage` to TypeScript;

No new changes to the component API!

### PR Checklist 🚀


packages/docs-page/index.tsx
```diff
@@ -48,7 +65,8 @@ export function DocsPageWrapper({
     <SearchProvider>
       <SearchBar
         product={name}
-        className={classNames({ [s.mobileSearch]: isMobile })}
+        // TODO; SearchBar doesn't accept `className`
+        // className={classNames({ [s.mobileSearch]: isMobile })}
```
thiskevinwang, line 69: Old bug?

packages/docs-page/index.tsx
```diff
@@ -127,7 +145,22 @@ export function DocsPageWrapper({
...
 }
 
-export default function DocsPage({
+export interface DocsPageProps {
+  product: { name: string; slug: string }
+  baseRoute: string
+  showEditPage: boolean
+  additionalComponents: MDXProviderComponentsProp
+  staticProps: {
+    mdxSource: any // TODO: import { MDXRemoteProps } from "next-mdx-remote"
```
thiskevinwang, line 154: I was thinking to do `Omit<MDXRemoteProps, "components">`

But we'll need to `export` it from https://github.com/hashicorp/next-mdx-remote/blob/main/src/index.tsx#L26

BRKalow: See this section of the docs, the `MDXRemoteSerializeResult` import should be the type we want here. https://github.com/hashicorp/next-mdx-remote#typescript

packages/docs-page/types.ts
```diff
@@ -0,0 +1,37 @@
...
+
+interface NavTree extends Node {
+  routes: (NavItem | ExternalItem | NavTree | Divider)[]
+}
+
+export interface NavData {
+  title: string
+  path: string
+  routes: NavTree
+}
```
BRKalow, line 37: I think these types are also repeated here: https://github.com/hashicorp/mktg-content-workflows/blob/main/shared/types.ts

I wonder if it would make sense to elevate these types to `@hashicorp/platform-types`. What do you think?

BRKalow: Actually, it's probably best to pull from here! https://github.com/hashicorp/react-components/blob/main/packages/docs-sidenav/types.ts

thiskevinwang: I'll delete this in favor of `import { NavData } from '@hashicorp/react-docs-sidenav/types'`

thiskevinwang: Addressed in https://github.com/hashicorp/react-components/pull/337/commits/a25c71a0b9b00298235785c04d4e10e095b7f63f

packages/docs-page/index.tsx
thiskevinwang: Addressed in https://github.com/hashicorp/react-components/pull/337/commits/a25c71a0b9b00298235785c04d4e10e095b7f63f

packages/docs-page/index.tsx
```diff
@@ -127,7 +144,26 @@ export function DocsPageWrapper({
...
+      page_title: string
+    }
+    currentPath: string
+    navData: NavData
+    githubFileUrl: string
+    versions: { name: string; label: string }[]
+  }
+}
+
+const DocsPage: ComponentType<DocsPageProps> = ({
```
BRKalow, line 166: As `DocsPage` doesn't accept children, I'm not sure it makes sense to use `ComponentType` here, which defines props as `PropsWithChildren<DocsPageProps>`. I would maybe just type props here:
```ts
export default function DocsPage({ ... }: DocsPageProps) {
```

ref: https://sourcegraph.com/github.com/DefinitelyTyped/DefinitelyTyped/-/blob/types/react/index.d.ts?L82:10

packages/docs-page/index.tsx
BRKalow: I think this might have been a casualty of the CSS modules migration. `SearchBar` here doesn't accept a `className`.

@thiskevinwang we should be update to update `SearchBar` to accept `className` and pass it on to the underlying search. https://github.com/hashicorp/react-components/blob/main/packages/docs-page/components/search-bar/index.jsx#L8

packages/docs-page/index.tsx
```diff
@@ -19,7 +22,20 @@ import LoadingSkeleton from './components/loading-skeleton'
...
+  currentPath: string
+  pageTitle: string
+  baseRoute: string
+  githubFileUrl: string
+  product: { name: string; slug: string }
+  showEditPage: boolean
+  versions: { name: string; label: string }[]
+}
+
+const DocsPageWrapper: ComponentType<DocsPageWrapperProps> = ({
```
BRKalow, line 38: We can refine this type further and use `FunctionComponent`. The `ComponentType` type also includes `ComponentClass`, which isn't the case here!

ref: https://sourcegraph.com/github.com/DefinitelyTyped/DefinitelyTyped/-/blob/types/react/index.d.ts?L82:10

BRKalow (commented):
Looking good!

packages/docs-page/components.ts
```diff
@@ -1,9 +1,10 @@
 import defaultMdxComponents from '@hashicorp/platform-docs-mdx'
+import { MDXProviderComponentsProp } from '@mdx-js/react'
 
 export default function generateComponents(
-  productName,
-  additionalComponents = {}
-) {
+  productName: string,
+  additionalComponents: MDXProviderComponentsProp = {}
+): Record<string, string> {
```
BRKalow, line 7: `defaultMdxComponents` returns a map of `string` to components I think 💭 

https://github.com/hashicorp/nextjs-scripts/blob/main/packages/docs-mdx/index.jsx#L11

packages/docs-page/index.tsx
thiskevinwang: Will update!

packages/docs-page/index.tsx
thiskevinwang: I'll opt for your suggestion, but will need to add an explicit return type `: JSX.Element`

We have a linting rule — `Missing return type on function. eslint@typescript-eslint/explicit-module-boundary-types` — that warns if you don't have an explicit return type.
- (Maybe we should re-assess this rule? Type-inference is pretty good, and sometimes explicit typing is prone to human error — 0 warnings, but incorrect type)

packages/docs-page/index.tsx
thiskevinwang: TIL!

packages/docs-page/components.ts
thiskevinwang: I'll do:

```ts
Record<string, (p: any) => JSX.Element>
```

packages/docs-page/index.tsx
BRKalow: Ah yeah, that rule! I think the value of that rule is debatable for react components, but it can be handy for other methods. We have had a few discussions about this rule in the past 💭  cc @hashicorp/web-platform

I think the return type should be `React.ReactElement`, FWIW. This matches the definition of the component types from `@types/react`.

packages/docs-page/components/search-bar/index.jsx
```diff
@@ -1,11 +1,12 @@
 import Search from '@hashicorp/react-search'
 import useIsMobile from '../../use-is-mobile'
 
-export default function SearchBar({ product }) {
+export default function SearchBar({ product, className }) {
   const isMobile = useIsMobile()
 
   return (
     <Search
+      className={className}
```
BRKalow, line 9: 💯

packages/docs-page/index.tsx
thiskevinwang: I was looking at `React.ReactElement` (from the Sourcegraph link) ... vs. `JSX.Element` (from type-hint/inference) 😅 

I'll update to `React.Element`!

BRKalow (approved):
🥇

PR #344, opened 2021-09-17
[Combobox] Move Formik to peerDeps

jmfury:
📝 To-Do
- [x] Run prerelease
- [x] Verify prerelease

---

## Description

`formik` must be a peer dependency so we do not install another copy of it. `<ComboboxField />` leverages the shared `FormikContext` via `useField`, so of course this needs to be using the same `formik`.

Hit this error when attempting to use `<ComboboxField />` over on another project. As an optional dependency, `npm` went ahead and installed a different `formik` local to this package (seen in the error).

image

### PR Checklist 🚀

### DIFF
--- a/.changeset/afraid-melons-collect.md
+++ b/.changeset/afraid-melons-collect.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-combobox': patch
+---
+
+Move formik to peerDependencies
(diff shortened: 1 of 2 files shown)


jmfury:
@BRKalow  Verified this fixes the issue, I do question how useful this `-Field` variant actually is. May come back later with a major to just remove it entirely, or provide more partial components that would complete the "story" e.g. built in error messages

BRKalow (approved):

PR #345, opened 2021-09-20
fix: button tertiary/neutral and hero light bg

BRKalow:
🎟️ [Asana Task]()

---

## Description

Updating to css modules on Boundary and running into a few small things with the CSS modules upgrade.

For button:
- Fixes the `tertiary`/`neutral` variant/brand combo to render with black text as expected (I think?)

For hero:
- Updates styles to add back missing light background styles, which updates the text and carousel elements to look better on light backgrounds.

### PR Checklist 🚀

### DIFF
--- a/packages/button/theme.module.css
+++ b/packages/button/theme.module.css
@@ -132,30 +132,42 @@
   }
 
   &.brand-neutral {
-    --background-color: transparent;
-    --border-color-hovered: transparent;
-    --border-color: transparent;
-    --padding-medium: 0;
-    --padding-small: 0;
-    --scrim-opacity-active: 0;
-    --scrim-opacity-hovered: 0;
     --text-color: var(--black);
 
     &.background-dark,
     &.background-brand {
       --text-color: var(--white);
     }
+  }
 
-    &[disabled] {
-      &.background-light,
-      &.background-brand-secondary {
-        --text-color: rgba(0, 0, 0, 0.3);
-      }
-      &.background-brand,
-      &.background-dark {
-        --text-color: rgba(255, 255, 255, 0.3);
-      }
+  &[disabled] {
+    &.background-light,
+    &.background-brand-secondary {
+      --text-color: rgba(0, 0, 0, 0.3);
     }
+    &.background-brand,
+    &.background-dark {
+      --text-color: rgba(255, 255, 255, 0.3);
+    }
+  }
+}
+
+/*
+ * Tertiary neutral variant
+ */
+.variant-tertiary-neutral {
+  --background-color: transparent;
+  --border-color-hovered: transparent;
+  --border-color: transparent;
+  --padding-medium: 0;
+  --padding-small: 0;
+  --scrim-opacity-active: 0;
+  --scrim-opacity-hovered: 0;
+  --text-color: var(--black);
+
+  &.background-dark,
+  &.background-brand {
+    --text-color: var(--white);
   }
 
   &[disabled] {
--- a/packages/button/theme.module.css
+++ b/packages/button/theme.module.css
@@ -132,39 +132,30 @@
   }
 
   &.brand-neutral {
-    &.background-dark {
+    --background-color: transparent;
+    --border-color-hovered: transparent;
+    --border-color: transparent;
+    --padding-medium: 0;
+    --padding-small: 0;
+    --scrim-opacity-active: 0;
+    --scrim-opacity-hovered: 0;
+    --text-color: var(--black);
+
+    &.background-dark,
+    &.background-brand {
       --text-color: var(--white);
     }
-  }
 
-  &[disabled] {
-    &.background-light,
-    &.background-brand-secondary {
-      --text-color: rgba(0, 0, 0, 0.3);
+    &[disabled] {
+      &.background-light,
+      &.background-brand-secondary {
+        --text-color: rgba(0, 0, 0, 0.3);
+      }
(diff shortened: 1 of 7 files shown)


packages/hero/style.module.css
```diff
@@ -227,3 +227,29 @@
     }
   }
 }
+
+.root:global(.light) {
```
BRKalow, line 231: Oops I think this needs to be not scoped to `.root` ideally...

BRKalow:
@zchsh I see now why we opted to have the video controls above the video element, the text overlain on the stripes isn't very readable
Screen Shot 2021-09-20 at 4 02 53 PM
:

packages/button/theme.module.css
```diff
@@ -132,40 +132,31 @@
...
+        --text-color: rgba(255, 255, 255, 0.3);
+      }
     }
   }
-}
-
-/*
- * Tertiary neutral variant
- */
-.variant-tertiary-neutral {
```
zchsh, line 155: I think this `.variant-tertiary-neutral` is still necessary to render that variant. `*`

I think the fix here might be simpler - eg [something like this](https://github.com/hashicorp/react-components/pull/295/commits/b552d23818c1adb6cf73dd383b25a8a0a251cd4d) (cherry-pick `b552d23 ` if you want) seems to produce [the expected sticker sheet appearance](https://react-components-storybook.vercel.app/?path=/story/components-button--themes).

---

`*` To be clear, `tertiary-neutral` is very duplicative of `{ variant: "tertiary", brand: "neutral" }`, it stemmed a design request to be able to conveniently ignore the brand colour in certain situations, as increasingly design wanted "neutral"-brand tertiary buttons alongside product-branded primary buttons. It was probably not the best thing to do, feels like it adds more confusion for the benefit of a very tiny amount of convenience.

packages/hero/carousel.module.css
```diff
@@ -147,3 +147,15 @@
     }
   }
 }
+
+:global(.light) .root {
```
zchsh, line 151: Thanks for catching this - facepalming over here cause I think I just ripped out the `light` theme and forgot to add it  back in 🤦 .

I think it might be preferable to keep CSS grouped by selector (some rationale for this in our Notion doc under `Code Style` / `CSS Modules`), rather than grouped by variation. One way to do this would be to replace the global `"light" | "dark"` class with `s[backgroundTheme]` classes on affected elements. Another way might be to only pass that class to the root, and set CSS custom properties that can then be hooked into further down in the cascade. [Here's a commit](https://github.com/hashicorp/react-components/pull/295/commits/cad9e8fc76cd391022cd1262546b42d5ade5797f) `cad9e8f ` you could cherry pick for the former option, and [a preview of the result](https://react-components-storybook.vercel.app/?path=/story/components-hero--light-theme).

zchsh (changes requested):
Big thanks for catching this and jumping in to help 💖  I think the removal of `.variant-tertiary-neutral` might cause problems - have some concrete suggestions in some hopefully easily cherry-pick-able commits 🙇

packages/button/theme.module.css
BRKalow: Oh shoot yeah I did not realize that one could use `tertiary-neutral` as the variant. I can add that back 👀. Thanks!

packages/hero/carousel.module.css
BRKalow: Okay so I'm a little bit on the fence about this, and maybe I've missed the boat on the discussion 😅 .

Looking at the commit you referenced, that to me seems like we are erring too much on the side of composability and re-usability at the component level. I think having something like the `theme` prop at the level of an exportable component makes sense, but if we're drilling it all the way down into the child components that are not exported or really re-usable it seems like we're losing some of the benefits of the cascade.

I'll try out putting the class at the root and setting properties, that seems like a reasonable approach!

It seems like there could be room for a `ThemeProvider` of some sort that helps with this 💭

BRKalow: Actually went with your commit for now, most in line with our current guidelines and consistency is a win 👍

zchsh: aack sorry i missed the comment earlier about the stripes being the reason for placement above - that's a great callout 💖 tbh not sure how to handle - looks like we have problems with the stripes on mobile anyways... but probably makes sense to prioritize desktop, so maybe we should add a `controlsFirst` prop or something, or maybe just always put them first?  🤔 augh i hate the number of options in this component ahah 😭

+1 on using the cascade by putting the class at the root and setting properties - or using a ThemeProvider if there are smaller components to drill into 💯 . I feel like i probably lean too heavily on just making things components sometimes, since I haven't used providers a ton yet (noting as context for why i went the direction i did and for future work ahah 😬 )

i probably also should have been more clear that's more just my hot take that i wrote down in Notion re: CSS, and not a discussion that's had a lot of activity, so super open to input and to changing our guidelines & recommendations there. I think your take of something more cascade-y for theming (whether through CSS modules or ThemeProvider of some sort) should be our first recommendation as a way to group styles by selector 👍  (also open to revising the recommendation to group styles by selector rather than state or theme, although that one does feel a little more solid IMO).

BRKalow:
@zchsh Thanks for the collab here! This should be good to go 🤞

zchsh (approved):
💖  Looks great, thanks for jumping in and handling this 🙇 

Also, I love all the changelog names 🍏👟🙅

PR #354, opened 2021-09-22
feat(docs-page): content-api calls

thiskevinwang:
🎟️ [Asana Task]()
🔍 [Preview Link](https://react-components-git-{branch-slug}-hashicorp.vercel.app)

---

## Description

For https://github.com/hashicorp/waypoint/pull/2344

This updates `docs-page` to request data from the [content api](https://github.com/hashicorp/mktg-content-workflows/)

Tests + TS Refactor:
- https://github.com/hashicorp/react-components/pull/360

### PR Checklist 🚀

Items in this checklist may not may not apply to your PR, but please consider each item carefully.

- [ ] Add Asana and Preview links above.
- [x] Conduct thorough self-review.
- [x] Add or update tests as appropriate.
- [ ] Conduct reasonable cross browser testing for both compatibility and responsive behavior (We have a [Sauce Labs](https://app.saucelabs.com/) account for this, if you don't have access, just ask!).
- [ ] Conduct reasonable accessibility review (use the [WAS](https://accessible.org/Web-Accessibility-Standards-WAS-2.pdf) as a guide or an [axe browser plugin](https://www.deque.com/axe/) until we establish more formal checks).
- [ ] Identify (in the description above) and document (add Asana tasks on [this board](https://app.asana.com/0/1100423001970639/list)) any technical debt that you're aware of, but are not addressing as part of this PR.

PR #360, opened 2021-09-28
refactor(docs-page): to TS; split up file

thiskevinwang:
[asana](https://app.asana.com/0/1199972122039519/1201012831626614)

# Description

- ♻️ This splits `docs-page/server.js` into isolated **TypeScript** files.
  - 🧪 This adds tests per file, with at least some basic coverage, like snapshots
- 🐛 This fixes some incorrect `moize` usage
- ⚡️ This improves types

This is meant to merge into: 
- https://github.com/hashicorp/react-components/pull/354


packages/docs-page/server/consts.ts
```diff
@@ -0,0 +1,10 @@
+export const ENABLE_VERSIONED_DOCS = process.env.ENABLE_VERSIONED_DOCS
+export const VERCEL_ENV = process.env.VERCEL_ENV
```
BRKalow, line 2: nit: I'm not super crazy about re-assigning env vars in a separate file like this. I find it harder to grok for usage and it's just another layer of indirection.

That said, I'm not willing to die on this hill and if you feel this is a better structure I'm okay with that. 😄

packages/docs-page/server/generate-static-props.ts
```diff
@@ -0,0 +1,154 @@
...
+} from '../content-api'
+import { ENABLE_VERSIONED_DOCS, DEFAULT_PARAM_ID } from './consts'
+
+import renderPageMdx from '../render-page-mdx'
+
+import { stripVersionFromPathParams, normalizeVersion } from '../util'
+
+import { resolveNavData } from './resolve-nav-data'
+
+const moiseOpts: Options = { isPromise: true, maxSize: Infinity }
```
BRKalow, line 16: good catch here

packages/docs-page/server/generate-static-props.ts
```diff
@@ -0,0 +1,154 @@
...
+        latestVersion = version
+      }
+
+      const displayValue = display ?? version
+
+      return {
+        name: isLatest ? 'latest' : version,
+        label: isLatest ? `${displayValue} (latest)` : displayValue,
+      }
+    })
```
BRKalow, line 71: seems like this could be a fair candidate to extract into a function that's easier to test

packages/docs-page/server/generate-static-props.ts
```diff
@@ -0,0 +1,154 @@
...
+          : normalizeVersion(versionFromPath)
+
+      const _fullPath = [
+        'doc',
+        currentVersionNormalized,
+        basePath,
+        ...pathParamsNoIndex,
+      ].join('/')
+
+      const documentPromise = fetchDocument(productSlug, _fullPath)
```
BRKalow, line 96:
```suggestion
      const fullPath = [
        'doc',
        currentVersionNormalized,
        basePath,
        ...pathParamsNoIndex,
      ].join('/')

      const documentPromise = fetchDocument(productSlug, fullPath)
```

BRKalow (commented):
🥇

packages/docs-page/server/generate-static-props.ts
thiskevinwang: 👍

packages/docs-page/server/consts.ts
thiskevinwang: I'm not tied to this! Happy to revert this change...

I'm more interested in lifting up environment variable usage into function args, so that the functions can be [more pure](https://en.wikipedia.org/wiki/Pure_function). Didn't get to this part yet though...

thiskevinwang: Something like 

```diff
- import { ENABLE_VERSIONED_DOCS, VERCEL_ENV, DEFAULT_PARAM_ID } from './consts'
+ import { DEFAULT_PARAM_ID } from './consts' 
```

Then in a function

```diff
- function generateStaticPaths(context) { ... }

+ const options = {
+  VERCEL_ENV: process.env.VERCEL_ENV,
+  ENABLE_VERSIONED_DOCS: process.env.ENABLE_VERSIONED_DOCS,
+ }
+ 
+ function generateStaticPaths(context, { VERCEL_ENV, ENABLE_VERSIONED_DOCS } = options) { ... }
```

packages/docs-page/server/generate-static-props.ts
thiskevinwang: need to change `moiseOpts` to `moizeOpts` 😂

packages/docs-page/server/generate-static-props.ts
thiskevinwang: extracted into `mapVersionList`

packages/docs-page/server/consts.ts
thiskevinwang: Updating this now... and removing the env vars from `consts.ts`

BRKalow:
Looking good! Could you test this with a canary release in a site? That would be full validation that the refactor is good, apart from your tests 😄

thiskevinwang:
Merging into #354 — will continue UAT there!

PR #363, opened 2021-09-29
fix(Hero): update alert text color

BRKalow:
🎟️ [Asana Task]()

---

## Description

Make the nested `Alert` in `Hero` have the opposite text color of what the `backgroundTheme` prop is specified as.

Screen Shot 2021-09-29 at 11 01 15 AM

### PR Checklist 🚀

### DIFF
--- a/packages/hero/index.js
+++ b/packages/hero/index.js
@@ -65,7 +65,7 @@ function Hero({ data, centered, gaPrefix, className, videoControlsTop }) {
               url={alert.url}
               tag={alert.tag}
               text={alert.text}
-              textColor={backgroundTheme}
+              textColor={backgroundTheme === 'light' ? 'dark' : 'light'}
               className={s.headlineAlert}
             />
           )}
--- a/.changeset/nice-ties-peel.md
+++ b/.changeset/nice-ties-peel.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-hero': patch
+---
+
+Adjust nested alert text color in hero to be the opposite of the backgroundTheme


packages/hero/index.js
```diff
@@ -65,7 +65,7 @@ function Hero({ data, centered, gaPrefix, className, videoControlsTop }) {
               url={alert.url}
               tag={alert.tag}
               text={alert.text}
-              textColor={backgroundTheme}
+              textColor={backgroundTheme === 'light' ? 'dark' : 'light'}
```
thiskevinwang, line 68: 2-second hand-wavy idea for the future

💭  We could pass `"invert"` that denotes `backgroundTheme === 'light' ? 'dark' : 'light'` logic
- basically, _the opposite of the current theme_

BRKalow: Pass `invert` to `Alert`? I wonder how we would implement that, as that would require `Alert` to be aware of its parent background color 🤔

thiskevinwang: ^ inspired by how Vercel handles this at the global level w/ `.invert-theme`
image

thiskevinwang (approved):

packages/hero/index.js
BRKalow: Gotcha. Yeah I guess we'd need a notion of a proper theme abstraction which we could then "invert." 😅 

Good idea!

thiskevinwang: Probably a farfetched thought — (I'm missing some pieces of how theming is handled)

jescalan (approved):
Thanks for cleaning this up - noticed this bug the other day and it drove me nuts

PR #370, opened 2021-09-30
test(version-select): coverage for URL sync

thiskevinwang:
🎟️ [-]()
🔍 [-](-)

---

## Description

Fixed by https://github.com/hashicorp/react-components/pull/369 (select-input 4.0.4)

This adds more test coverage

Smoke test: https://waypoint-none31bq2-hashicorp.vercel.app/commands which is using `select-input 4.0.4` under the hood

### PR Checklist 🚀

### DIFF
--- a/.changeset/weak-chairs-judge.md
+++ b/.changeset/weak-chairs-judge.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-version-select': patch
+---
+
+Adds additional test coverage. Depends on https://github.com/hashicorp/react-components/pull/369
--- a/.changeset/weak-chairs-judge.md
+++ b/.changeset/weak-chairs-judge.md
@@ -1,5 +1,0 @@
----
-'@hashicorp/react-version-select': patch
----
-
-Adds additional test coverage. Depends on https://github.com/hashicorp/react-components/pull/369
--- a/packages/version-select/index.test.tsx
+++ b/packages/version-select/index.test.tsx
@@ -58,7 +58,7 @@ describe('<VersionSelect />', () => {
     const { getByRole } = render(<VersionSelect {...defaultProps} />)
 
     const button = getByRole('button')
-    expect(button).toHaveTextContent('latest')
+    expect(button).toHaveTextContent('v0.5.x (latest)')
   })
 
   it('should display version from the URL', () => {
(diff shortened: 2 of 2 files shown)


packages/version-select/index.test.tsx
```diff
@@ -76,6 +76,33 @@ describe('<VersionSelect />', () => {
...
+    const combobox = getByRole('combobox')
+    userEvent.click(combobox)
+
+    const options = getAllByRole('option')
+    userEvent.click(options[index])
+
+    history.back()
+
+    const button = getByRole('button')
+    expect(button).toHaveTextContent('v0.5.x (latest)')
```
BRKalow, line 103: What's the value before the back? Given that we're mocking the router I'm not sure what exactly we're asserting here 🤔

thiskevinwang: It would be `"v0.4.x"` — I can add an assertion before to be safe

BRKalow: I don't think this is gonna work how we expect here, cause the router is mocked so the URL isn't going to be changing.

BRKalow: I think the only way to test this would be to change the return value of the router mock and then force the component to re-render.

In which case the value of this test case is a little dubious as we already have test cases to assert we are getting the version from the URL, and the behavior of `SelectInput` when `value` changes should be tested there.

thiskevinwang: I believe you're right 🤔

thiskevinwang: I have a test that is failing for `select-input 4.0.3` but working on `select-input 4.0.4`

```ts
  it('should display the correct version after browser-back', () => {
    useRouterMock.mockImplementation(() => {
      return {
        asPath: '/commands',
        push: mockPush,
      } as unknown as Router
    })

    const { getByRole, rerender } = render(<VersionSelect {...defaultProps} />)

    expect(getByRole('button')).toHaveTextContent('v0.5.x (latest)')

    // simulate navigating: latest -> v0.4.x
    useRouterMock.mockImplementation(() => {
      return {
        asPath: '/commands/v0.4.x',
        push: mockPush,
      } as unknown as Router
    })
    rerender(<VersionSelect {...defaultProps} />)

    expect(getByRole('button')).toHaveTextContent('v0.4.x')

   // simulate navigating: v0.4.x -> latest
    useRouterMock.mockImplementation(() => {
      return {
        asPath: '/commands/',
        push: mockPush,
      } as unknown as Router
    })
    rerender(<VersionSelect {...defaultProps} />)

    expect(getByRole('button')).toHaveTextContent('v0.5.x (latest)')
  })
```

.changeset/weak-chairs-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hashicorp/react-version-select': patch
+---
+
+Adds additional test coverage. Depends on https://github.com/hashicorp/react-components/pull/369
```
BRKalow, line 5: If we're just adding tests, we can omit the changeset!

BRKalow (commented):
👍

.changeset/weak-chairs-judge.md
thiskevinwang: TIL! Removing...

BRKalow (approved):
👏

packages/version-select/index.test.tsx
```diff
@@ -54,26 +54,39 @@ describe('<VersionSelect />', () => {
     expect(options).toHaveLength(VERSIONS.length)
   })
 
-  it('should display the latest version', () => {
-    const { getByRole } = render(<VersionSelect {...defaultProps} />)
+  it('should display the correct version after router events', () => {
```
thiskevinwang, line 57:
```suggestion
  it('should display the correct version from the URL', () => {
```

PR #381, opened 2021-10-04
feat(subnav): add support for dark theme

zchsh:
🎟️ [Asana Task](https://app.asana.com/0/1100423001970639/1201105487825905/f)
🔍 [Preview Link](https://react-components-git-zssubnav-dark-theme-hashicorp.vercel.app)
📖 [Storybook Preview](https://react-components-storybook.vercel.app/?path=/story/components-subnav--cloud-hashicorp)
🐣 [Canary Test Preview](https://cloud-hashicorp-com-git-zsupgrade-subnav-hashicorp.vercel.app) via https://github.com/hashicorp/cloud.hashicorp.com/pull/192

---

This PR adds support for a `theme` prop to `@hashicorp/react-subnav`. The intent here is to enable [a dark theme, as implemented in cloud.hashicorp.com](https://github.com/hashicorp/cloud.hashicorp.com/blob/main/components/DarkSubnav/DarkSubnav.module.css) (see [preview](https://cloud.hashicorp.com)) without using potentially brittle overrides (which are also now much harder to write at the project level due to our use of CSS modules).

### PR Checklist 🚀


packages/subnav/theme.module.css
```diff
@@ -0,0 +1,43 @@
+/* Theme settings. Note: also consumed in /partials */
```
BRKalow, line 1: ❤️

packages/subnav/index.js
```diff
@@ -2,6 +2,7 @@ import classNames from 'classnames'
 import useProductMeta from '@hashicorp/platform-product-meta'
 import useNavRef from './helpers/useNavRef'
 import s from './style.module.css'
+import themeStyles from './theme.module.css'
```
zchsh, line 5: Took a theming approach similar to that in `code-block`. While I am certainly not confident this will be the best approach to theming long-term (in fact I'm almost sure we can come up with something more extensible for product situations), this felt like a decent balance between using our existing tooling and conventions (CSS modules & custom properties) while splitting out theme properties.

My hot take is I don't think the theming is something we should aim to "solve" with this PR, really just trying to get this to work and be easy to transition to whatever theming approach we go with longer term. Very open to suggestions here, of course 🙇

packages/subnav/index.js
```diff
@@ -115,7 +119,7 @@ function Subnav({ className, ...restProps }) {
   return (
     <nav
       ref={wrapperRef}
-      className={classNames(s.root, className, {
+      className={classNames(s.root, themeStyles[theme], className, {
```
zchsh, line 122: MOST of the theming is done via CSS custom properties - that comes from this class here 👍

packages/subnav/index.js
```diff
@@ -125,6 +129,7 @@ function Subnav({ className, ...restProps }) {
         product={product}
         hasOverflow={hasOverflow}
         isSticky={isSticky}
+        theme={theme}
```
zchsh, line 132: SOME of the theming, namely just the theme passed to `button`'s `theme.background`, has to be passed through the `CtaLinks` component. This is a part that doesn't feel ideal long term (Context probably better?) but felt sustainable and reasonable given the scope & intent of this PR.

packages/subnav/partials/CtaLinks/github-stars-link/style.module.css
```diff
@@ -1,10 +1,10 @@
 .root {
-  --stars-border-color: rgba(29, 30, 35, 0.2);
-  --scrim-opacity: 0;
+  --local-scrim-color: transparent;
+  --local-divider-color: var(--border-color); /* from theme.module.css */
```
zchsh, line 5: These `--local` vars make it possible to group styles fully by element. Without them I think we'd have to group by state, which might be okay in this specific situation, but quickly gets messy so I think should be avoided (per CSS guidelines in Notion - though as always, open to discussion on those as well).

packages/subnav/partials/CtaLinks/icons/github-logo.js
```diff
@@ -5,6 +5,7 @@ function GithubIcon(props) {
     <svg width={16} height={16} viewBox="0 0 16 16" {...props}>
       <path
         fillRule="evenodd"
+        fill="#000000"
```
zchsh, line 8: Added this as it's usually present in icon exports, and makes it possible to target the icon's fill in a less brittle way.

packages/subnav/partials/CtaLinks/index.js
```diff
@@ -41,6 +40,8 @@ function CtaLinks(props) {
             theme={{
               brand: product,
               variant: isLastButton ? 'primary' : 'secondary',
+              background: theme,
+              ...link.theme /* allow theme overrides via the ctaLinks array */,
```
zchsh, line 44: Added this in to remove a related override in cloud.hashicorp.com used to emulate a `secondary` button. Consumers can now add a `link.theme` attribute to set button themes, if desired.

packages/subnav/theme.module.css
```diff
@@ -0,0 +1,47 @@
...
+  --gh-stars-icon-color: var(--black);
+  --gh-stars-scrim-color: rgba(0, 0, 0, 0.1);
+
+  /* menu items */
+  --menu-item-text-color: var(--black);
+  --menu-item-underline-color: var(--black);
+  --menu-item-heading-text-color: var(--gray-3);
+}
+
+.dark {
```
zchsh, line 28: I've riffed off the existing implementation at cloud.hashicorp.com. Have also changed the modals to match the dark theme of the bar to avoid a checkerboard look.

Not 100% sure if this matches design's intent - I found what I think is the [reference Figma file](https://www.figma.com/file/qy89mZUl1HHFZ1d3Pc4s29/HCP-.com-Experience-Landing-Page?node-id=2605%3A9358) but could not find deets on the nav theming. So, went with what I _think_ would match design intent.

BRKalow:
Nice work here! I think the only thing that I see in the HCP preview is that the logo svg paths have a black fill. 💭

zchsh:
@BRKalow ugh, aack... yeah the CSS there is pretty brittle. Will switch to loading in the white logos proper 👍 Extra weight but probably worth the stability.

zchsh:
Had some late-night-still-pretty-sick brain there but https://cloud-hashicorp-com-git-zsupgrade-subnav-hashicorp.vercel.app should be fixed now with all the new logos 👍 

5pa9nx

> Note: also tested out all the light / dark and product combos on [the Storybook preview](https://react-components-storybook.vercel.app/?path=/story/components-subnav--cloud-hashicorp) to make sure they're all loading.

BRKalow (approved):
This all looks good to me, thanks for tackling this! Nice balance of get-it-out-the-door and forward-thinking 🙇

PR #391, opened 2021-10-12
Add permalink spacing handling to subnav when its sticky

jescalan:
🎟️ [Asana Task]()

---

## Description

When the subnav is in sticky mode, and you have a permalink on the page, the headline ends up scrolling underneath the menu. This little css thingy prevents that from happening. It used to be implemented on each of our documentation sites hanging off the global `.g-subnav` but that no longer works because of css modules, and honestly it should have been part of this component anyway since I can't really imagine a scenario where you would want your permalinks to be hidden by this component, so I think this should do the trick. Tested on vault docs website and it worked wonderfully.

### PR Checklist 🚀

### DIFF
--- a/.changeset/sixty-humans-fix.md
+++ b/.changeset/sixty-humans-fix.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-subnav': patch
+---
+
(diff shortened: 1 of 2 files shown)


BRKalow (approved):
Nice, thanks. I had seen this during the upgrade and was wondering what exactly it was for.

PR #402, opened 2021-10-19
fix(open-api-page): patch issue with sidenav link, versioning

zchsh:
🎟️ [Asana Task](https://app.asana.com/0/1100423001970639/1201221479692275/f)
🔍 [Preview Link](https://react-components-git-zsopen-api-page-patch-hashicorp.vercel.app/components/openapipage)

---

This PR patches an issue in `@hashicorp/react-open-api-page` where for single-service use cases, the sidebar item was not linked correctly.

This PR also fixes a versioning blunder - previously we published `2.0.0` outside of the usual changsets workflow. This PR bumps `open-api-page` to `2.0.0` and marks as a patch release. This is intended to avoid a potential future conflict with the accidentally published `2.0.0` version.

Have proven this out via pre-release:
- https://github.com/hashicorp/cloud.hashicorp.com/pull/211

### PR Checklist 🚀

Items in this checklist may not may not apply to your PR, but please consider each item carefully.

- [x] Add Asana and Preview links above.
- [x] Conduct thorough self-review.
- [x] Add or update tests as appropriate.
- [x] Conduct reasonable cross browser testing for both compatibility and responsive behavior (We have a [Sauce Labs](https://app.saucelabs.com/) account for this, if you don't have access, just ask!).
- [x] Conduct reasonable accessibility review (use the [WAS](https://accessible.org/Web-Accessibility-Standards-WAS-2.pdf) as a guide or an [axe browser plugin](https://www.deque.com/axe/) until we establish more formal checks).
- [x] Identify (in the description above) and document (add Asana tasks on [this board](https://app.asana.com/0/1100423001970639/list)) any technical debt that you're aware of, but are not addressing as part of this PR.

PR #404, opened 2021-10-19
fix(open-api-page): patch issue when there are multiple services

zchsh:
🎟️ [Asana Task](https://app.asana.com/0/1100423001970639/1201221479692275/f)
🔍 [Preview Link](https://react-components-git-zsopen-api-patch-for-boundary-hashicorp.vercel.app)

---

This PR makes a patch to `@hashicorp/react-open-api-page` missed in #402 , which supports the Boundary use case where there are multiple "service" groups (and corresponding pages) where operations are rendered.

Have proven this out via pre-release in:
- Packer: https://github.com/hashicorp/cloud.hashicorp.com/pull/211
- Boundary: https://github.com/hashicorp/boundary/pull/1519

### PR Checklist 🚀

### DIFF
--- a/.changeset/tough-pandas-pay.md
+++ b/.changeset/tough-pandas-pay.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-open-api-page': patch
+---
+
+Patches issue with multi-service page rendering.


packages/open-api-page/server.js
```diff
@@ -35,26 +35,25 @@ function getPropsForPage(schema, params) {
   // rather than a specific operation-slug-based path as it would otherwise
   if (isSingleService) navData[0].path = ''
   // If there's no "page" param, then this is the landing page
-  const isLanding = !params || !params.page || params.page.length == 0
   const currentPath = params && params.page ? params.page.join('/') : ''
-
+  const isLanding = !params || !params.page || currentPath == ''
```
zchsh, line 39: Refactored the logic with an early return for landing pages to make things more clear.

PR #414, opened 2021-10-29
feat: use next/script in ConsentManager

BRKalow:
🎟️ [Asana Task]()

---

## Description

Refactoring ConsentManager to use `next/script` under the hood, as well as some refactoring to a functional component and hooks.

* Adds a new `strategy` attribute on services, which we can specify to pass along a desired loading strategy to `next/script`
* Refactors the top-level `ConsentManager` component to TS and away from a class component
* Cleans up tests and internal logic

### PR Checklist 🚀


packages/consent-manager/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+        />
+      )}
+      <SegmentScript
+        preferences={preferences}
+        writeKey={props.segmentWriteKey}
+      />
+      <CustomScripts
+        preferences={preferences}
+        services={props.additionalServices}
+      />
```
BRKalow, line 120: This replaces the previous `loadAnalytics` method. It's a bit more declarative and easier to trace now, these components will inject the necessary scripts based on the user's preferences.

packages/consent-manager/loader/index.tsx
```diff
@@ -50,10 +57,10 @@ export default function createConsentManager({
   }
 
   // next we build the config objct, kicking it off with the default values
-  let config: ConsentManagerConfig = {
-    version: 4,
-    companyName: 'HashiCorp',
-    privacyPolicyLink: 'https://www.hashicorp.com/privacy',
+  let config: ConsentManagerProps = {
```
BRKalow, line 60: Standardizing on shared types

mwickett (approved):
This looks really nice. Great to see an older component like this get cleaned up, and get improvements.

I've given the code a high-level look, and tested in the Swingset preview to see that scripts are in fact being loaded when consent is given.

jescalan (commented):
Looks great!

jescalan (approved):

thiskevinwang (approved):
LGTM!

Nit: Maybe next >=11 should be specified as a peer dep? ...since `next/script` is an [11+](https://nextjs.org/blog/next-11) API

Which also makes me curious: Is this a major or minor change?

BRKalow:
@thiskevinwang Yeah good call I'll add next to peer deps. I think strictly by the book this could be a major for the reason you said, but since all of our stuff is on next 11 and we control the apps this runs in it is reasonable for it to be a minor.

PR #424, opened 2021-11-15
feat(search): expose algolia config as props, fallback to env

zchsh:
🎟️ [Asana Task](https://app.asana.com/0/0/1201348169048582/f)
🔍 [Preview Link](https://react-components-git-zssearch-flexibility-hashicorp.vercel.app)

This PR exposes `react-search`'s Algolia configuration as an `algoliaConfig` prop that can be passed to `<SearchProvider />`. This change has been proven out in https://github.com/hashicorp/dev-portal/pull/36. 

---

This change is intended to be non-breaking: the new `algoliaConfig` prop is optional, and `react-search` will continue to use `.env` variables if the `algoliaConfig` prop is not available. If neither form of configuration is available, we'll continue to throw an error (as we did previously if `.env` vars were not present).

This PR also modifies `react-docs-page` to expose the same configuration variable, which is passed to our use of `react-search` in `docs-page`. As with `react-search`, this is a non-breaking feature addition to `docs-page`.

---

### PR Checklist 🚀

### DIFF
--- a/packages/docs-page/index.test.tsx
+++ b/packages/docs-page/index.test.tsx
@@ -30,7 +30,7 @@ describe('<DocsPage />', () => {
   beforeEach(() => {
     jest.clearAllMocks()
     useRouterMock.mockImplementation(() => routerMock)
-    headMock.mockImplementation(({ children }) => children)
+    headMock.mockImplementation(({ children }) => <>{children}</>)
   })
 
   it('passes `title`, `description`, and `siteName` correctly to <HashiHead>', () => {
--- a/packages/docs-page/index.test.tsx
+++ b/packages/docs-page/index.test.tsx
@@ -87,12 +87,8 @@ describe('<DocsPage />', () => {
       return <strong>Text in custom component</strong>
     }
     const additionalComponents = { CustomComponent }
-    const {
-      mdxSource,
-      frontMatter,
-    } = await renderPageMdx(
-      "## Heading Two\n\nHere's a paragraph of content.\n\n<CustomComponent />",
-      { productName: 'Terraform', additionalComponents }
+    const { mdxSource, frontMatter } = await renderPageMdx(
+      "## Heading Two\n\nHere's a paragraph of content.\n\n<CustomComponent />"
     )
     render(
       <DocsPage
@@ -112,12 +108,8 @@ describe('<DocsPage />', () => {
   })
 
   it('initializes jump to section UI if there is an h1 and two or more h2s', async () => {
-    const {
-      mdxSource,
-      frontMatter,
-    } = await renderPageMdx(
-      "---\n\npage_title: Test Title\ndescription: Test description\n---\n\n# Heading One\n\nAn intro paragraph.\n\n## Heading Two\n\nHere's a paragraph of content.\n\n## Here a second heading\n\nAnd another paragraph.",
-      { productName: 'Terraform' }
+    const { mdxSource, frontMatter } = await renderPageMdx(
+      "---\n\npage_title: Test Title\ndescription: Test description\n---\n\n# Heading One\n\nAn intro paragraph.\n\n## Heading Two\n\nHere's a paragraph of content.\n\n## Here a second heading\n\nAnd another paragraph."
     )
     render(
       <DocsPage
--- a/packages/search/index.test.js
+++ b/packages/search/index.test.js
@@ -22,13 +22,13 @@ function setupEnv() {
   process.env = { ...originalEnv } //  copy original environment
 }
 
-function restoreEnv() {
+function teardownEnv() {
   process.env = originalEnv //  restore original environment
 }
 
 describe('<Search />', () => {
   beforeAll(setupEnv)
-  afterAll(restoreEnv)
+  afterAll(teardownEnv)
   beforeEach(setDummyVars)
 
   it('should pass className to the root element', () => {
@@ -58,7 +58,7 @@ describe('<Search />', () => {
 
 describe('<Hits />', () => {
   beforeAll(setupEnv)
-  afterAll(restoreEnv)
+  afterAll(teardownEnv)
   beforeEach(setDummyVars)
 
   it('should display no results with invalid input', () => {
@@ -129,7 +129,7 @@ describe('<SearchProvider /> configuration', () => {
     //  Restore console.error for further tests
     global.console.error.mockRestore()
     // Restore the original .env for further tests
-    restoreEnv()
+    teardownEnv()
   })
 
   it('should allow mixed use of Algolia config props and .env', () => {
@@ -153,13 +153,13 @@ describe('<SearchProvider /> configuration', () => {
     expect(screen.getByText(testIndexName)).toBeInTheDocument()
 
     // Restore the original .env for further tests
-    restoreEnv()
+    teardownEnv()
   })
 })
 
 describe('<SearchProvider /> context', () => {
   beforeAll(setupEnv)
-  afterAll(restoreEnv)
+  afterAll(teardownEnv)
   beforeEach(setDummyVars)
 
   it('should provide a context object', () => {
@@ -207,7 +207,7 @@ describe('<SearchProvider /> context', () => {
 
 describe('search tools', () => {
   beforeAll(setupEnv)
-  afterAll(restoreEnv)
+  afterAll(teardownEnv)
 
   let algoliaConfig = {}
 
--- a/packages/search/index.test.js
+++ b/packages/search/index.test.js
@@ -9,10 +9,6 @@ import Search, {
 import { HitsComponent } from './hits'
 import { indexContent } from './tools'
 
-const renderWithProvider = (ui) => {
-  return render(<SearchProvider>{ui}</SearchProvider>)
-}
-
 const originalEnv = process.env
 
 const setDummyVars = () => {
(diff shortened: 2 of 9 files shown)


packages/docs-page/index.test.tsx
```diff
@@ -30,7 +30,7 @@ describe('<DocsPage />', () => {
   beforeEach(() => {
     jest.clearAllMocks()
     useRouterMock.mockImplementation(() => routerMock)
-    headMock.mockImplementation(({ children }) => children)
+    headMock.mockImplementation(({ children }) => <>{children}</>)
```
zchsh, line 33: I was getting a TypeScript warning here, so updated to avoid it.

packages/search/index.test.js
```diff
@@ -9,10 +9,6 @@ import Search, {
 import { HitsComponent } from './hits'
 import { indexContent } from './tools'
 
-const renderWithProvider = (ui) => {
```
zchsh, line 12: I removed this helper, as I wanted to configure `<SearchProvider />` with props in specific tests.

packages/search/index.test.js
```diff
@@ -63,30 +62,102 @@ describe('<Hits />', () => {
...
-
     const resultsEl = screen.getByRole('listbox')
     expect(resultsEl).toHaveAttribute('id', SEARCH_RESULTS_ID)
     const hitItems = screen.getAllByTestId('hit-item')
     expect(hitItems.length).toBe(2)
   })
 })
 
-describe('<SearchProvider />', () => {
+describe('<SearchProvider /> configuration', () => {
```
zchsh, line 92: I've added tests to ensure the configuration functionality works as expected.

packages/search/provider.tsx
```diff
@@ -0,0 +1,115 @@
+import React, { useMemo, useState, createContext, useContext } from 'react'
+import algoliaSearch, { SearchClient } from 'algoliasearch'
+import searchInsights from 'search-insights'
+import { AlgoliaConfigObject } from './types'
+
+const SearchContext = createContext(null)
```
zchsh, line 6: Converted `provider` to TypeScript. TypeScript warned about `createContext` needing a default value, so I've used `null`.

packages/search/provider.tsx
```diff
@@ -0,0 +1,115 @@
...
+    },
+  }
+
+  function initAlgoliaInsights() {
+    searchInsights('init', { appId, apiKey })
+  }
+
+  function logClick(hit) {
+    // --------------
+    // TODO: I was unable to resolve the TypeScript error here.
```
zchsh, line 79: I ran into this TypeScript issue, timeboxed myself a few minutes to try to resolve it, and wasn't able to. If anyone has any insight into how to resolve this issue, would love any help 🙇 

I don't love having this messy comment, but I also don't love sinking a ton of time into "fixing" TS errors in code that I'm not really trying to touch.... and given the code here is identical I think perhaps it's acceptable for now? cc @BRKalow the TypeScript wizard 🧙 (also no pressure here, I'm happy to try to dive in further and try to fix this if needed).

packages/search/types.ts
```diff
@@ -0,0 +1,8 @@
+export interface AlgoliaConfigObject {
```
zchsh, line 1: Added a types file as `react-docs-page` uses the same interface for its `algoliaConfig` prop. This seems to be a common pattern in our component library, that we're starting to adopt, but wanted to call it out as I'm not sure how "blessed" it is yet.

zchsh:
@jescalan Here's the fix for `react-search` we discussed briefly - I've got it proven out at https://github.com/hashicorp/dev-portal/pull/36 (preview links on that PR show separate Boundary and Waypoint search working as expected).

@BRKalow tagged you here cause I ran into a TypeScript issue that I'm _hoping_ is something simple that I'm just not familiar with yet... but if it's not, or if you're slammed, feel free to untagged yourself and I can dive deeper to try to figure it out 👯

jescalan (approved):
This is great 👏

packages/search/provider.tsx
BRKalow: @zchsh It wasn't you! Looks like the types are wrong from the package. I pushed up a minor change. The types are fixed in v2.1.0, we are currently on v1.8.0. Feel free to take a stab at upgrading if you're feeling _dangerous_.

packages/search/provider.tsx
```diff
@@ -0,0 +1,105 @@
...
+      value={{
+        client,
+        indexName,
+        initAlgoliaInsights,
+        isCancelled,
+        logClick,
+        query,
+        setCancelled,
+        setQuery,
+      }}
```
BRKalow, line 100: When creating provider values, I like to try and memoize them to ensure that they aren't causing unnecessary renders. In this case, it looks like we can declare `client`, `initAlgoliaInsights`, and `logClick` all within a single `useMemo`:

```tsx
const contextValue = useMemo(() => ({
  client: {
    search: (requests) => { ... }
  },
  initAlgoliaInsights: () => { ... },
  logClick: (hit) => { ... },
}), [ ... ])

// ...

<SearchContext.Provider value={contextValue}>
  {children}
</SearchContext.Provider>
```

zchsh: oh nice, this makes sense, I was pretty rushed with making things "not global" to get them in scope of the props. ty 💖

packages/search/provider.tsx
zchsh: Alright, I ran into what feels like an out-of-scope issue with the upgrade:

Based on the README for v2.x.x, the only breaking change is that `useCookie` will default to `false` rather than `true`. Based on [this note on cookies](https://github.com/algolia/search-insights.js#cookie-usage), and [the `useCookie` entry in the options documentation table](https://github.com/algolia/search-insights.js#2-initialize-the-library), we'll need to explicitly set `useCookie` to `true` in order to track _any_ search events (otherwise they'll error). I think this would affect search events such as `logClick`, [as used in react-search hits](https://github.com/hashicorp/react-components/blob/03e3ed2549d79f632673f702db1a63748063bb95/packages/search/hits.js#L178) and therefore in docs-page, or such as [`logConversion` in learn](https://github.com/hashicorp/learn/blob/4a2458b69844a8eb542baeba008fb5d302794713/lib/hooks/log-search-conversion.js#L15)).

In other words, it seems that as part of the upgrade, and perhaps even as part of our current use, we should consider hooking into `ConsentManager` before calling eg `logClick` or `logConversion`.

So, going to hold off on the upgrade, and have added a task to our web-platform board to discuss 👍

packages/search/provider.tsx
zchsh: ✅ Updated to `useMemo` on `client`, `initAlgoliaInsights`, and `logClick` as described 👍

zchsh:
@BRKalow Tagged you for a quick confirmation that the `useMemo` approach for `client` etc looks as expected - I did not memoize the _entire_ context value, just the Algolia client parts, and wanted to check whether that seemed correct 🙇

packages/search/provider.tsx
```diff
@@ -0,0 +1,113 @@
...
+        client,
+        indexName,
+        initAlgoliaInsights,
+        logClick,
+        // State
+        query,
+        setQuery,
+        isCancelled,
+        setCancelled,
+      }}
```
BRKalow, line 108: So the key here is that the full context value object is memoized, otherwise it'll change referentially on each render. We'll want to include the entire object as declared here in the `useMemo` above and pass the result directly to the provider.

```tsx
const contextValue = useMemo(() => ({ ... })

<SearchContext.Provider value={contextValue}>
```

BRKalow: Sorry I wasn't clear the first time around 😅

zchsh: Can do - I am curious to try to figure something out while we're at it, cause I'm not as familiar with `useMemo`.

If we're declaring the `algoliaClient` instance like `const algoliaClient = algoliaSearch(appId, apiKey)` in the `useMemo`, and `query` is a dependency of that `useMemo`, then we'll be re-running those declarations every time `query` changes, ie every time someone types something in the search bar, right?

eg

```
// Memo-ize the full context value,
// (is there any performance concern with re-declaring `algoliaClient` etc
// every time the `query` string changes?)
const contextValue = useMemo(() => (
   const algoliaClient = algoliaSearch(appId, apiKey)
    return {
      client: { ... }.
      initAlgoliaInsights: () => searchInsights('init', { appId, apiKey }),
      logClick: (hit) => { ... },
      indexName,
      query,
      setQuery,
      isCancelled,
      setCancelled,
    }
), [ apiKey, appId, indexName, query, setQuery, isCancelled, setCancelled ])
```

Maybe this is inconsequential performance wise, but wondering if it makes any sense to do eg:

```
// Memo-ize client context values, as these should only
// change and cause re-renders when Algolia configuration changes
 const { client, initAlgoliaInsights, logClick } = useMemo(() => {
    const algoliaClient = algoliaSearch(appId, apiKey)
    return {
      client: { ... }.
      initAlgoliaInsights: () => searchInsights('init', { appId, apiKey }),
      logClick: (hit) => { ... },
    }
  }, [apiKey, appId, indexName])

// Memo-ize the full context value,
// if this is not done then the `value` object would change referentially on each render
const contextValue = useMemo(() => (
    return { client, initAlgoliaInsights, logClick, indexName, query, setQuery, isCancelled, setCancelled }
), [ client, initAlgoliaInsights, logClick, indexName, query, setQuery, isCancelled, setCancelled ])
```

(Or maybe another way to put this is I don't currently understand whether the above two blocks are really that different, but would like to!)

BRKalow: > If we're declaring the algoliaClient instance like const algoliaClient = algoliaSearch(appId, apiKey) in the useMemo, and query is a dependency of that useMemo, then we'll be re-running those declarations every time query changes, ie every time someone types something in the search bar, right?

That's right!

I missed that query was part of the value here, whoops. To me, your second case makes sense as those three values are relatively stable and static, whereas the others are less-so with the inclusion of `query`.

zchsh: awesome possum! Thanks for the thorough review and patience here, the objects-change-referentially thing is definitely one of those things I am really not as familiar with so kind of a big blind spot for me sometimes 😅 I started https://justjavascript.com at some point, and this thread is making me wanna go back and work all the way through it 🙇

BRKalow: It's definitely an interesting concept that can have a big impact on React performance. Here's some quick links:
* https://dev.to/tylerthehaas/referential-equality-in-react-127h
* https://kentcdodds.com/blog/how-to-optimize-your-context-value

BRKalow (approved):

PR #427, opened 2021-11-16
chore(docs-sidenav): add missing key prop

thiskevinwang:
🎟️ [Asana Task](https://app.asana.com/0/1100423001970639/1201378049617080)

---

## Description

This adds a missing `key` prop

### PR Checklist 🚀

### DIFF
--- a/packages/docs-sidenav/index.js
+++ b/packages/docs-sidenav/index.js
@@ -174,7 +174,8 @@ function NavTree({ baseRoute, content }) {
 
     // Headings
     if (item.heading) {
-      return <h3>{item.heading}</h3>
+      // eslint-disable-next-line react/no-array-index-key
+      return <h3 key={idx}>{item.heading}</h3>
     }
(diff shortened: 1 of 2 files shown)


packages/docs-sidenav/index.js
```diff
@@ -174,7 +174,8 @@ function NavTree({ baseRoute, content }) {
 
     // Headings
     if (item.heading) {
-      return <h3>{item.heading}</h3>
+      // eslint-disable-next-line react/no-array-index-key
+      return <h3 key={idx}>{item.heading}</h3>
```
BRKalow, line 178:
```suggestion
      return <h3 key={item.heading + idx}>{item.heading}</h3>
```

BRKalow: Since we have a string value available :+1:

BRKalow (approved):
Noice good catch

PR #454, opened 2021-12-06
Fix a bug in older firefox versions

jescalan:
🎟️ [Asana Task]()

---

## Description

Potentially fixes https://github.com/hashicorp/vault/issues/13350

Tested in saucelabs, breaking down [the shorthand](https://developer.mozilla.org/en-US/docs/Web/CSS/flex) seems to resolve the issue.

### PR Checklist 🚀

### DIFF
--- a/packages/docs-page/style.module.css
+++ b/packages/docs-page/style.module.css
@@ -104,7 +104,9 @@ which is rendered into docs-sidenav */
   padding: 0 1rem;
 
   @media (--desktop-viewports) {
-    flex: 1;
+    flex-grow: 1;
+    flex-shrink: 1;
+    flex-basis: 0;
     margin: 0;
     padding-left: 2rem;
     padding-right: 0;
--- a/packages/docs-page/style.module.css
+++ b/packages/docs-page/style.module.css
@@ -104,8 +104,7 @@ which is rendered into docs-sidenav */
   padding: 0 1rem;
 
   @media (--desktop-viewports) {
-    flex-grow: 1;
-    flex-shrink: 1;
+    flex: 1 1 auto;
     margin: 0;
     padding-left: 2rem;
     padding-right: 0;
--- a/packages/docs-page/style.module.css
+++ b/packages/docs-page/style.module.css
@@ -106,7 +106,6 @@ which is rendered into docs-sidenav */
   @media (--desktop-viewports) {
     flex-grow: 1;
     flex-shrink: 1;
-    flex-basis: 0;
     margin: 0;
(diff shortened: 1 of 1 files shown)


dstaley:
Applying the expanded form doesn't seem to resolve the issue for me in 78esr

image

jescalan:
@dstaley could you try again but with dropping the `flex-basis`? I just adjusted the PR for this 😬

dstaley:
> @dstaley could you try again but with dropping the `flex-basis`? I just adjusted the PR for this 😬

Negative ghost rider.

image

jescalan:
@dstaley ah you also need to disable `flex` declaration attached directly to the selector for it to kick in 😁

zchsh:
Tested on SauceLabs, I think I'm seeing the same thing as @dstaley . When I add `min-width: 0` to `#inner` it seems to resolve the issue though 👍 I think this has to do with how the default `min-width: auto` works in `flex` contexts (https://drafts.csswg.org/css-flexbox/#min-size-auto) but am unclear on the details (have just ran into this before!).

zchsh:
CleanShot 2021-12-07 at 15 22 49@2x

zchsh:
Edit: looked into this a bit more and found a simpler fix. It seems that the defaults for `flex-basis` have changed over time. Explicitly setting  `flex: 1 1 auto` seems to resolve the issue as well.

CleanShot 2021-12-07 at 15 27 53@2x

packages/docs-page/style.module.css
```diff
@@ -104,7 +104,8 @@ which is rendered into docs-sidenav */
   padding: 0 1rem;
 
   @media (--desktop-viewports) {
-    flex: 1;
+    flex-grow: 1;
+    flex-shrink: 1;
```
zchsh, line 108: I'm honestly having trouble understanding all the nuance both of flex-basis and how it has changed over the years (https://developer.mozilla.org/en-US/docs/Web/CSS/flex-basis#values). But taking a more pragmatic approach, the CSS below seems to work both in FireFox 78 and in newer browsers:

```suggestion
    flex: 1 1 auto;
```

As an alternate solution, this also seems to work:

```suggestion
    flex: 1 1;
    min-width: 0;
```

zchsh (changes requested):

jescalan:
Sounds good, applied the suggestion!

zchsh (approved):

zchsh:
💖 ☝️ Sorry for the delay 😬

PR #462, opened 2021-12-08
fix: only set aria-activedescendant to an active ID

dstaley:
🎟️ [Asana Task]()

---

## Description

The `aria-activedescendant` must be set to an element ID as defined by the [spec](https://www.w3.org/TR/2017/REC-wai-aria-1.1-20171214/#aria-activedescendant). Setting it to an empty string renders the prop in the HTML, whereas setting it to `undefined` results in the property not being set on the element.

### PR Checklist 🚀

### DIFF
--- a/.changeset/five-comics-perform.md
+++ b/.changeset/five-comics-perform.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-search': patch
+---
+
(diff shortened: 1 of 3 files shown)


alexcarpenter (approved):
LGTM 👍🏼

PR #481, opened 2022-01-05
Enable `strict` mode

dstaley:
🎟️ [Asana Task](https://app.asana.com/0/1100423001970639/1201500876162997/f)

---

## Description

This PR enables `strict` in our `tsconfig`, in addition to fixing the compiler errors that arise as a result of the change. The only `strict` option we currently cannot enable is `noImplicitAny`, so that's been disabled.

For the most part, I preferred asserting that the current usage was valid rather than introduce a significant number of logic changes across several packages. In some instances, however, I felt it relatively safe to modify the logic to be valid with the types as written.

One of the largest set of changes came from our `ref` usage. The types for refs are not exactly the easiest things to work with, so if something looks funky please let me know and we can take a closer look. Everything passes the compiler, but that doesn't necessarily mean it's "correct"!

## Todo

- [ ] Add typechecking to CI
  - [ ] Do we want GitHub Actions or CircleCI?
  - [ ] Do we want to require a passing typecheck in order to commit to `main`?

### PR Checklist 🚀

### DIFF
--- a/packages/docs-page/server/index.ts
+++ b/packages/docs-page/server/index.ts
@@ -117,7 +117,7 @@ export function generateStaticPaths({
   const loader = new FileSystemLoader({
     navDataFile,
     localContentDir,
-    product: product?.slug,
+    product: product?.slug ?? '',
     paramId,
   })
 
--- a/packages/docs-page/server/index.ts
+++ b/packages/docs-page/server/index.ts
@@ -99,7 +99,7 @@ export interface GenerateStaticPathsContext {
   /**
    * @example { name: 'Waypoint', slug: 'waypoint' }
    */
-  product?: { name: string; slug: string }
+  product: { name: string; slug: string }
   /** @example 'docs' */
   basePath?: string
 }
@@ -117,7 +117,7 @@ export function generateStaticPaths({
   const loader = new FileSystemLoader({
     navDataFile,
     localContentDir,
-    product: product?.slug ?? '',
+    product: product.slug,
     paramId,
   })
 
--- a/packages/tabs/partials/tab-triggers/index.tsx
+++ b/packages/tabs/partials/tab-triggers/index.tsx
@@ -27,10 +27,10 @@ function TabTriggers({
   className,
   theme = 'light',
 }: TabTriggersProps): React.ReactElement {
-  const overflowBaseRef = useRef(null)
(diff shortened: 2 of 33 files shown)


BRKalow:
This seems like a great baseline to me. Acknowledging that usage of `!` _can_ be unsafe, but that everything is functionally unchanged with this, I think that's okay. Your ref typings look correct to me.

> Do we want GitHub Actions or CircleCI?

GitHub Actions!

> Do we want to require a passing typecheck in order to commit to main?

I would say yes!

packages/tabs/partials/tab-triggers/index.tsx
```diff
@@ -61,7 +63,7 @@ function TabTriggers({
    */
   useEffect(() => {
     // Determine which arrows to show
-    const { scrollLeft, scrollWidth, offsetWidth } = scrollRef.current
+    const { scrollLeft, scrollWidth, offsetWidth } = scrollRef.current!
```
BRKalow, line 66: This potentially feels like a scenario where we might want to check that `scrollRef.current` is defined, but as it was like this before I don't think it's critical.

packages/docs-page/server/index.ts
```diff
@@ -117,7 +117,7 @@ export function generateStaticPaths({
   const loader = new FileSystemLoader({
     navDataFile,
     localContentDir,
-    product: product?.slug,
+    product: product?.slug ?? '',
```
thiskevinwang, line 120:
```suggestion
    product: product.slug,
```

thiskevinwang: And, I think [product isn't supposed to be optional](https://github.com/hashicorp/react-components/pull/481/files#diff-99829c81bc27331948908774fd93aa5a7e75b42d97274484f95d2c11d133025fR102) 🤔 

@BRKalow?

thiskevinwang (commented):
Just 1 comment; And I'm a big fan of strict null checks! 🌶️ 🎉

packages/docs-page/server/index.ts
BRKalow: Right, it probably shouldn't be optional as we need it for the edit this page link.

dstaley: 👍

BRKalow (approved):

PR #484, opened 2022-01-10
chore: Hide VersionAlert on DocsPage

thiskevinwang:
- chore(DocsPage): hide VersionAlert
- chore: changeset

🎟️ [Asana Task](https://app.asana.com/0/1100423001970639/1201619796458628/f)

---

## Description

This updates `DocsPage` to **not** display the `VersionAlert` if the version in the path happens to be the "latest" version

Depends on 
- https://github.com/hashicorp/mktg-content-workflows/pull/103
- manually update DB data

### PR Checklist 🚀

### DIFF
--- a/packages/docs-page/index.tsx
+++ b/packages/docs-page/index.tsx
@@ -22,7 +22,7 @@ import LoadingSkeleton from './components/loading-skeleton'
 import useIsMobile from './use-is-mobile'
 import s from './style.module.css'
 
-interface DocsPageWrapperProps {
+interface DocsPageInnerProps {
   canonicalUrl: string
   description: string
   navData: NavData
@@ -38,7 +38,7 @@ interface DocsPageWrapperProps {
   algoliaConfig?: AlgoliaConfigObject
 }
 
-export const DocsPageWrapper: FunctionComponent<DocsPageWrapperProps> = ({
+export const DocsPageInner: FunctionComponent<DocsPageInnerProps> = ({
   canonicalUrl,
   children,
   description,
@@ -56,6 +56,9 @@ export const DocsPageWrapper: FunctionComponent<DocsPageWrapperProps> = ({
   const isMobile = useIsMobile()
   const { asPath } = useRouter()
   const versionInPath = getVersionFromPath(asPath)
+
+  const versionIsLatest =
+    versionInPath && versions?.[0]?.label.includes(versionInPath)
 
   // TEMPORARY (https://app.asana.com/0/1100423001970639/1160656182754009)
   // activates the "jump to section" feature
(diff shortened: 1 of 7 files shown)


packages/docs-page/index.tsx
```diff
@@ -57,6 +57,9 @@ export const DocsPageWrapper: FunctionComponent<DocsPageWrapperProps> = ({
   const { asPath } = useRouter()
   const versionInPath = getVersionFromPath(asPath)
 
+  const versionIsLatest =
+    versionInPath && versions?.[0]?.label.includes(versionInPath)
+
```
thiskevinwang, line 62: 💭 This logic is pretty brittle...

thiskevinwang: This ticket will update https://app.asana.com/0/1100423001970639/1201619796458626/f all products to have a generic (x-patch) latest version

BRKalow: Almost feels like this would be made easier with some kind of "activeVersion" data available from a context or something similar. That would also help persisting the version across different sections.

thiskevinwang: @BRKalow Yea, i've been thinking about that, and I'm pretty on board!

I see the `activeVersion` context helping to de-dupe client-side state-wrangling and props drilling.
- 💭 it would still be derived from the `next/router` right (I still understand that as being a more comprehensive source of truth)

I'd also throw in _`activeLane`_ in there (groupings of _subPaths_ — docs/api | plugins — that should be versioned together in the UI)

---

Tangent to your comment, but related to this PR. I just had an idea to return _more_ data from server code
- https://github.com/hashicorp/react-components/blob/main/packages/docs-page/server/loaders/remote-content.ts#L57-L76

The current implementation is kind of causing this client-side [pain point / fragility](https://github.com/hashicorp/react-components/pull/484/files/6e2febb0c696a3604a9edcf545e8533c5fc64a46#diff-3a175ef635fbea4834759b73520c2e10cbef9d293bc962b8860e85aed42b10c5R60-R61)

BRKalow: Returning an `isLatest` attribute from the server would be a good starting point for sure. Yeah ultimately we would need to derive the `activeVersion` from the router I think, but like you said avoid prop drilling and de-duping state seems like a win. We can definitely defer it, or maybe use the thoughts here to draft up a more comprehensive task for future us.

thiskevinwang: I'm going to address the server code bit in this PR to unblock consul versioned docs

But for the context bit, I was thinking it'd make sense to do async w/ the consul/packer versioned docs rollout

thiskevinwang: > We can definitely defer it, or maybe use the thoughts here to draft up a more comprehensive task for future us.

I'll make a ticket right now!

thiskevinwang:
Here's a screenshot of `Consul` on 
`"@hashicorp/react-docs-page": "^14.14.0-canary-2022011213255",`

### Versioned url; Latest

image

BRKalow (approved):
rad

PR #488, opened 2022-01-14
fix(react-content): Fix mismatch between table width and child element widths

BRKalow:
🎟️ [Asana Task]()

---

## Description

The adjustment to add x-axis overflow scrolling on tables caused some weird behavior:

Screen Shot 2022-01-14 at 9 47 30 AM
(see https://www.consul.io/commands/join)

Largely this appears to be because of the change to `display: block` so that we could apply a `max-width`, but this is causing other issues.

I've changed the display to `inline-block`, which appears to fix the width issue while still allowing us to retain the x-axis scrolling.

I had to remove the `margin-bottom` because margin collapsing does not apply to `inline-block` elements, per spec.

After:
Screen Shot 2022-01-14 at 9 50 00 AM
Screen Shot 2022-01-14 at 9 49 43 AM

### PR Checklist 🚀

---
  - https://app.asana.com/0/0/1201661195283485

### DIFF
--- a/.changeset/spicy-geckos-type.md
+++ b/.changeset/spicy-geckos-type.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-content': patch
+---
+
+Update table styles so that the top-level table width and child table element widths do not get mismatched, while still retaining the max-width and x-axis scroll.
--- a/packages/content/styles/table.module.css
+++ b/packages/content/styles/table.module.css
@@ -1,13 +1,15 @@
 .table {
   & table {
-    display: inline-block;
+    display: block;
     overflow-x: auto;
     border-collapse: collapse;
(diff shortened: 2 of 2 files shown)


thiskevinwang (approved):
LGTM! Curious to hear @alexcarpenter's input here too

BRKalow:
@zchsh Yeah that's a fair question! I think it should be possible with this styling by just specifying `width: 100%;` 👍

alexcarpenter (approved):
ah sorry I missed small tables with the recent update. `max-content` seems like the ideal fix vs `inline-block`. Originally proposed `max-content` but the [caniuse](https://caniuse.com/?search=max-content) seemed to inconsistent but testing in both FF and Safari seemed to work just fine.

PR #493, opened 2022-01-20
feat: add error-view component

zchsh:
🎟️ [Asana Task](https://app.asana.com/0/1199972122039519/1201534814309539)
🔍 [Preview Link](https://react-components-git-zsadd-error-view-hashicorp.vercel.app/components/errorview)

---

## Description

This PR adds a `@hashicorp/react-error-view` component.

Have tested a canary release of this component in https://github.com/hashicorp/vagrant/pull/12651.

### PR Checklist 🚀

Items in this checklist may not may not apply to your PR, but please consider each item carefully.

- [x] Add Asana and Preview links above.
- [x] Conduct thorough self-review.
- [x] Add or update tests as appropriate.
- [x] Conduct reasonable cross browser testing for both compatibility and responsive behavior (We have a [Sauce Labs](https://app.saucelabs.com/) account for this, if you don't have access, just ask!).
- [x] Conduct reasonable accessibility review (use the [WAS](https://accessible.org/Web-Accessibility-Standards-WAS-2.pdf) as a guide or an [axe browser plugin](https://www.deque.com/axe/) until we establish more formal checks).
- [x] Identify (in the description above) and document (add Asana tasks on [this board](https://app.asana.com/0/1100423001970639/list)) any technical debt that you're aware of, but are not addressing as part of this PR.

PR #497, opened 2022-01-25
adds versioned error view

BRKalow:
🎟️ [Asana Task](https://app.asana.com/0/1100423001970639/1201630522524202/f)

---

## Description

**Note: this PR is targeting @zchsh's branch to add the error-view package. [The PR](https://github.com/hashicorp/react-components/pull/493)**

Adds a specific error view for use on docs sites with versioned docs. It's effectively a full, custom implementation of `pages/_error.tsx`. Implementation looks something like this:

```tsx
import { VersionedErrorPage } from '@hashicorp/react-error-view/versioned-view'

export default VersionedErrorPage
```

Screen Shot 2022-01-25 at 12 18 48 PM

### PR Checklist 🚀

### DIFF
--- a/packages/error-view/versioned-view.tsx
+++ b/packages/error-view/versioned-view.tsx
@@ -2,12 +2,12 @@ import React from 'react'
 import { useRouter } from 'next/router'
 import Link from 'next/link'
 import useErrorPageAnalytics from './use-error-page-analytics'
-import ErrorPage from '.'
+import ErrorPage, { ErrorPageProps } from '.'
 import s from './style.module.css'
 
 const versionPattern = /\/(?<version>v\d+[.]\d+[.](\d+|x))/
 
-function VersionNotFound({ version }): React.ReactElement {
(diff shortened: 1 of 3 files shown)


packages/error-view/versioned-view.tsx
```diff
@@ -0,0 +1,78 @@
...
+          <a className={s.link}>← Go back to Documentation</a>
+        </Link>
+      </p>
+    </div>
+  )
+}
+
+/**
+ * A Standalone error page component intended to be used on sites with versioned docs support.
+ * Exports a full, custom implementation of an pages/_error.tsx component.
```
BRKalow, line 40: I opted to use `pages/_error` here instead of `pages/404` as the page renders differently depending on the URL, and so static generation of the 404 page doesn't really make a lot of sense IMO. The alternative is to SSR it one way and do some client-side munging to render it differently after the initial hydration, at which point why not just SSR?

zchsh (approved):
✅ This looks great to me! On board with using `pages/_error` 👍

PR #502, opened 2022-01-28
Add `use404Redirects` hook to error-view exports

dstaley:
🎟️ [Asana Task]()

---

## Description

This PR adds a new hook called `use404Redirects` to the exports of `@hashicorp/react-error-view`. When this hook is used on a 404 page, it checks to see if the current URL results in a 200 status from the server with a `pathname` different to the current value. If so, it performs a client-side redirect to that new `pathname`.

This hook allows us to cover the edge-case where the Next Router _thinks_ a URL can be resolved solely on the client, and thus doesn't run any redirects.

### PR Checklist 🚀

### DIFF
--- a/packages/error-view/__tests__/use-404-redirects.test.tsx
+++ b/packages/error-view/__tests__/use-404-redirects.test.tsx
@@ -1,0 +1,83 @@
+import nock from 'nock'
+import { render, waitFor } from '@testing-library/react'
+import * as Router from 'next/router'
+import use404Redirects from '../use-404-redirects'
+
+function HookTestComponent() {
+  use404Redirects()
+  return <div>Testing the use404Redirects hook...</div>
+}
+
+const useRouter = jest.spyOn(Router, 'useRouter')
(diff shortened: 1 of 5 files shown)


zchsh (approved):
This looks great to me! 💯

packages/error-view/__tests__/use-404-redirects.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+import use404Redirects from '../use-404-redirects'
+
+function HookTestComponent() {
+  use404Redirects()
+  return <div>Testing the use404Redirects hook...</div>
+}
+
+const useRouter = jest.spyOn(Router, 'useRouter')
+
+describe('use404Redirects', () => {
```
BRKalow, line 13: nice tests!

BRKalow (approved):
Looks good! Just wanna highlight that this is a really clever  approach, nicely done! 🏆

PR #512, opened 2022-02-23
feat: `<Badge />` component

alexcarpenter:
🎨 [Figma Component](https://www.figma.com/file/7cYgDM618stjYUHDqAfRec/Components?node-id=211%3A719)
🔍 [Preview Link](https://react-components-git-acbadge-component-hashicorp.vercel.app/components/badge)

---

## Description

This PR aims to introduce a new `<Badge />` component in line with the new Web Pattern Library Figma component implementation.

### PR Checklist 🚀


packages/badge/index.tsx
```diff
@@ -0,0 +1,21 @@
+import { Products } from '@hashicorp/platform-product-meta'
+import classnames from 'classnames'
+import s from './style.module.css'
+
+interface BadgeProps {
+  children: string
```
dstaley, line 6: I'm curious if you considered using a prop for the text value instead, like `<Badge title="" />`?

packages/badge/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+  variant?: 'primary' | 'secondary'
+}
+
+export default function Badge({
+  children,
+  variant = 'primary',
+  theme = 'neutral',
+}: BadgeProps) {
+  return (
+    <span className={classnames(s.badge, s[theme], s[variant])}>
```
BRKalow, line 17: Did we look at using the product meta styles for this? https://github.com/hashicorp/web-platform-packages/tree/main/packages/product-meta

dstaley, line 17: I'll defer to @ashleemboyer, but I vaguely recall badges like this benefiting from some aria props like `role="status"`. That pertains more to badges that are being updated as a result of user action which doesn't seem to be the intent here, but I'm curious if we should go ahead and add those attributes now to support that use case in the future.

dstaley (approved):
😍

BRKalow (commented):
Looks good!

packages/badge/style.module.css
```diff
@@ -0,0 +1,135 @@
+.badge {
+  background-color: var(--background-color, var(--black));
+  color: var(--color, var(--white));
+  border-radius: 4px;
+  display: block;
+  composes: g-type-label from global;
```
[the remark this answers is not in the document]
alexcarpenter: 👍🏼 https://github.com/hashicorp/react-components/pull/512/commits/78db41fe7e90870d8cf37761d2b1766d00f889c5

packages/badge/index.tsx
ashleemboyer: Setting `role="status"` has side effects, like setting `aria-live` to `polite`, so I'd say let's hold off on adding ARIA until we know more about how this component will be used. But my gut take is that if the text in this badge is static / doesn't update while a user is looking at it, then `role="status"` shouldn't be used.

Source: https://www.w3.org/TR/wai-aria/#status

packages/badge/index.tsx
alexcarpenter: I wanted to propose we use children because it naturally follows how you'd use an html element. For example you want to place text inside a paragraph `<p>Test</p>` vs `<p title="Test" />`. I've found it to be confusing the use of text or title between our components. [Button](https://react-components.vercel.app/components/button) expects `title`, where [CallToAction](https://react-components.vercel.app/components/calltoaction) links expect `text`.

jmfury: +1 - That was what I had in t[his initial iteration on a project](https://hashicorp-www-next-git-jmhomepage-promo-section-hashicorp.vercel.app/swingset/components/badge), I wondered if there was a strong reason _for_ using `children`

packages/badge/index.tsx
alexcarpenter: I did, but the theme support more than our product colors. We'd be loading unnecessary variable declarations that are not documented to use within the Figma component implementation.

BRKalow: The product meta theme classes are more than likely already in the document though, so it shouldn't be anything net-new. We could  use a combination of the styles from product meta as well as ones which are for the unique themes that `Badge` supports. Which unnecessary variable declarations are we worried about?

packages/badge/index.tsx
EnMod: +100! The "pure HTML" feel of wrapping an element's content is very intuitive, and as a bonus it promotes composition. I dig this!

packages/badge/index.tsx
alexcarpenter: @BRKalow Here is an update making use of productStyles https://github.com/hashicorp/react-components/pull/512/commits/282709c8e36a8166ca5df4627c8f8d4fa4689187

BRKalow: oh snap

BRKalow (approved):
👏

PR #522, opened 2022-03-02
Remove fragment from `<UseCases />` 

jmfury:
🎟️ [Asana Task]()

---

## Description

We are cleaning up our DatoCMS models, and [plan to delete `sbcUseCasesSection`](https://github.com/hashicorp/react-components/pull/522/files#diff-2e5597201fb9bbcf837a4af09a067ac01a6a7f63c12ee936278e4fbdd69b8e33L1) - the only projects using this component are our `.io` websites, but it does not seem they are using the fragment within any `rivet` GraphQL requests. This is likely to be a safe upgrade for all consumers, but is a major version bump given the breaking changes.

### PR Checklist 🚀


zchsh (approved):
💖 [I love to see it](https://app.asana.com/0/1199912664053769/1199521916370261/f)

PR #532, opened 2022-03-08
feat(code-block): left-align tabs, fix valid children issue

zchsh:
🎟️ [Asana Task - Alignment](https://app.asana.com/0/1100423001970639/1201866381269498/f)
🎟️ [Asana Task - Children validation fix](https://app.asana.com/0/1100423001970639/1201881376116200/f)
🔍 [Preview Link](https://react-components-git-zscode-tabs-alignment-hashicorp.vercel.app/components/codeblock)

---

## Description

This PR...

This PR also fixes a bug in `CodeTabs` related to validation of children elements, which prevented some use in JSX. With this fix in place, `CodeTabs` should correctly validate the use of `CodeBlock` children in JSX.

### PR Checklist 🚀

---
  - https://app.asana.com/0/0/1201881376116200
  - https://app.asana.com/0/0/1201866381269498

### DIFF
--- a/packages/code-block/docs.mdx
+++ b/packages/code-block/docs.mdx
@@ -275,17 +275,41 @@ function hello() {
 
 The `CodeTabs` component, which is available in Learn and Docs MDX contexts, allows authors to create multi-snippet tabbed code blocks. `<CodeTabs>` is a wrapper component around child code blocks. Each child must be either a fenced code block, or must be a `<CodeBlockConfig>` which wraps a fenced code block.
 
+By default, the tab options are shown at the left of the component. When space is limited, tabs will automatically collapse into a dropdown menu. When a `heading` is provided to `CodeTabs`, the heading will be shown at left, and the tabs (or dropdown menu) will be shown at right.
+
+#### `CodeTabs` Playground
+
+<LiveComponent>{`
+<CodeTabs theme="dark">
+
+<CodeBlock
+  options={{ showClipboard: true, filename: 'hello-world.js' }}
+  language="javascript"
+  code={\`<span class="token console class-name">console</span><span class="token punctuation">.</span><span class="token method function property-access">log</span><span class="token punctuation">(</span><span class="token string">'hello world'</span><span class="token punctuation">)</span>\`}
+/>
+
+<CodeBlock
+  options={{ showClipboard: true }}
+  language="go"
+  code={\`<span class="token keyword">package</span> main<!-- -->
+<!-- -->
+<span class="token keyword">import</span> <span class="token string">"fmt"</span>
+<!-- -->
+<span class="token keyword">func</span> <span class="token function">main</span><span class="token punctuation">(</span><span class="token punctuation">)</span> <span class="token punctuation">{</span>
+<!-- -->  fmt<span class="token punctuation">.</span><span class="token function">Println</span><span class="token punctuation">(</span><span class="token string">"hello world"</span><span class="token punctuation">)</span>
+<span class="token punctuation">}</span>\`}
+/>
+
+</CodeTabs>
+`}</ LiveComponent>
+
 #### When to use `CodeTabs`
 
-Tabs should only be used when switching between similar content types such as:
+Tabs should only be used when switching between equivalent approaches to a particular task or workflow. Common examples include alternate configuration type, programming, languages, or operating systems:
 
 - Windows / macOS / Linux
 - HCL / JSON
 - JavaScript / BASH
-
-Essentially either similar languages or OSs, or more broadly, equivalent approaches to a particular task or workflow.
-
-Where space is limited, tabs will automatically collapse into a dropdown menu on the right side.
 
 #### With fenced code child tabs
 
--- a/packages/code-block/partials/code-tabs/index.js
+++ b/packages/code-block/partials/code-tabs/index.js
@@ -66,6 +66,9 @@ function CodeTabs({ children, heading, className, tabs, theme = 'dark' }) {
   // gather labels
   const tabLabels = parsedTabs.map((t) => t.label)
 
+  // if we have a heading, we align tabs right (and make some style tweaks)
+  const hasHeading = Boolean(heading)
+
   return (
     <div
       className={classNames(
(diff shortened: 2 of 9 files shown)


packages/code-block/docs.mdx
```diff
@@ -275,18 +275,42 @@ function hello() {
 
 The `CodeTabs` component, which is available in Learn and Docs MDX contexts, allows authors to create multi-snippet tabbed code blocks. `<CodeTabs>` is a wrapper component around child code blocks. Each child must be either a fenced code block, or must be a `<CodeBlockConfig>` which wraps a fenced code block.
 
+By default, the tab options are shown at the left of the component. When space is limited, tabs will automatically collapse into a dropdown menu. When a `heading` is provided to `CodeTabs`, the heading will be shown at left, and the tabs (or dropdown menu) will be shown at right.
+
+#### `CodeTabs` Playground
```
zchsh, line 280: Added a section in Swingset to help demonstrate the new alignment feature.

packages/code-block/docs.mdx
```diff
@@ -275,18 +275,42 @@ function hello() {
...
+<span class="token punctuation">}</span>\`}
+/>
+
+</CodeTabs>
+`}</ LiveComponent>
+
 #### When to use `CodeTabs`
 
-Tabs should only be used when switching between similar content types such as:
+Tabs should only be used when switching between equivalent approaches to a particular task or workflow. Common examples include alternate configuration type, programming, languages, or operating systems:
```
zchsh, line 308: Rephrased this a bit to get to the point and then show examples - it felt like the previous text I'd written was a bit unclear.

packages/code-block/partials/code-tabs/index.js
```diff
@@ -23,14 +25,31 @@ function CodeTabs({ children, heading, className, tabs, theme = 'dark' }) {
...
-    // For function components in JSX, ie CodeBlock and CodeBlockConfig,
-    // tabChild.type is a function whose name we need
-    if (typeof tabChild.type == 'function') type = tabChild.type.name
-    // In MDX contexts, the component type is captured in mdxType
-    if (typeof tabChild.props.mdxType === 'string')
+    // For JSX primitives, the type is captured by the type property
+    if (typeof tabChild.type == 'string' || typeof tabChild.type == 'number') {
+      type = tabChild.type
+      // For function components, accept CodeBlock or CodeBlockConfig.
+    } else if (typeof tabChild.type === 'function') {
```
zchsh, line 32: This builds on the approach Bryce suggested to maintain the existing behaviour of flagging all invalid types in a an error message. Due to the minification concern, there's a bit of repetition here to make sure the intended names make it into the error message; my thinking is that's a decent tradeoff for having clearer, easier-to-debug error messages.

Have also made a small style change to use braces for the `if` blocks 👍

packages/code-block/partials/code-tabs/index.test.js
```diff
@@ -1,14 +1,15 @@
...
+import CodeBlockConfig from '../code-block-config'
 import CodeTabsProvider from '../../provider'
 
 // getAllByText is needed, as we sometimes render dummy elements
 // before the main, displayed element, eg for overflow detection
 // and for simpler copy-to-clipboard logic
 function getSecondByText(text) {
   /* eslint-disable-next-line no-unused-vars */
-  const [firstElem, secondElem] = screen.getAllByText(text)
+  const [_firstElem, secondElem] = screen.getAllByText(text)
```
zchsh, line 12: Nit, this is an unused variable, prefixing with `_` makes VSCode happy (at least in my case).

packages/code-block/partials/code-tabs/index.test.js
```diff
@@ -25,6 +26,35 @@ it('should accept a className prop, and place it on the root element', () => {
   expect(container.firstChild.className).toContain(customClassName)
 })
 
+it('should accept CodeBlockConfig or pre elements as children', () => {
```
zchsh, line 29: Added a test to assert that the changes made don't affect the existing behaviour.

packages/code-block/partials/code-tabs/partials/tabs-as-tabs/index.js
```diff
@@ -2,16 +2,20 @@ import React from 'react'
 import classNames from 'classnames'
 import s from './style.module.css'
 
-function TabsAsTabs({ tabLabels, activeTabIdx, setActiveTabIdx }) {
+function TabsAsTabs({ tabLabels, activeTabIdx, setActiveTabIdx, hasHeading }) {
   return (
-    <div className={s.tabsAsTabs}>
+    <div className={classNames(s.tabsAsTabs, { [s.hasHeading]: hasHeading })}>
```
zchsh, line 7: To achieve left-alignment when a `heading` is not present, but maintain the existing right-alignment when a `heading` is present, we add some styling for `TabsAsTabs` (and the space-limited alternate `TabsAsDropdowns`) which is applied conditionally based on the presence of a `heading`.

zchsh:
@ashleemboyer This fixes the `CodeTabs` issue (encountered on the Install views), and updates tabs to be left-aligned (unless a `heading` is provided). I've proven this out in https://github.com/hashicorp/dev-portal/pull/169. Let me know what you think! 🙇‍♂️

packages/code-block/partials/code-tabs/index.test.js
ashleemboyer: If you want I think you could update this to not name any variable at all:

```suggestion
  const [, secondElem] = screen.getAllByText(text)
```

zchsh: nice, TIL!

ashleemboyer (approved):
LGTM! 🥇👏🏻

zchsh:
Wrestling with the canary release workflow, it seems to be hanging for some reason (usually takes < 1 min and taking WAY longer). Have run into this before, there's no reason why the code change in that suggestion should affect it, I think it's just flaky. Trying one more time, but going to merge even if it fails (cause it's some strange issue with the workflow, not with that latest commit).

PR #547, opened 2022-03-25
Update CodeBlock Copy button style and icon

ashleemboyer:
🎟️ [Asana Task](https://app.asana.com/0/1201979821989355/1201959792992259/f)
🔍 [Preview Link](https://react-components-git-ambupdate-code-block-copy-button-hashicorp.vercel.app/components/codeblock)

---

## Description

- Adds `@hashicorp/flight-icons` package
- Updates Copy button style based on the design screenshot below
- Uses some flight icons for the default, success, and failed copy states

<details>
  <summary>New button states</summary>
  <a href="https://www.figma.com/file/VD7ahvXuXWJApeGnhbW4hv/?node-id=7606%3A75239">Figma link</a>
  CleanShot 2022-06-02 at 14 00 04@2x
</details>

### PR Checklist 🚀


package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@hashicorp/react-components",
   "version": "0.0.0",
   "dependencies": {
+    "@hashicorp/flight-icons": "^2.1.1",
```
ashleemboyer, line 5: Not sure if this needs to go in the CodeBlock package.json or not?

packages/code-block/partials/clipboard-button/style.module.css
```diff
@@ -1,42 +1,83 @@
-/*
-Note: Expects CSS properties from 
-code-block/theme-(dark|light).module.css
-to be present.
-*/
-
 .button {
```
ashleemboyer, line 1: TODO: double check if the font-weight is correct

packages/code-block/partials/clipboard-button/style.module.css
```diff
@@ -1,42 +1,84 @@
...
+    border: 3px solid transparent;
+    bottom: -4px;
+    box-sizing: border-box;
+    content: '';
+    left: -4px;
+    position: absolute;
+    right: -4px;
+    top: -4px;
+    z-index: 1;
   }
```
alexcarpenter, line 46: 💭 Is this same implementation not achievable using outline and outline-offset vs introducing a pseudo element?

alexcarpenter (commented):
Definitely outside of this PR and something we can look into in the future, but curious if we can keep the focus on the button vs sending it back to the document on click. Not sure why that happens.

packages/code-block/partials/clipboard-button/style.module.css
```diff
@@ -1,42 +1,84 @@
...
+  border-radius: 5px;
+  border-style: solid;
+  border-width: var(--border-width);
+  color: var(--token-color-foreground-high-contrast);
   display: flex;
-  padding: 0 16px;
-  line-height: 32px;
+  font-family: system-ui, -apple-system, blinkmacsystemfont, 'Segoe UI', roboto,
+    helvetica, arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji',
+    'Segoe UI Symbol';
```
zchsh, line 26: 🎨 (maybe 🚧 blocking?) This is interesting - we use `code-block` pretty heavily on marketing websites, and of course on the existing dot-ios which do _not_ use the system font stack... I think this might stand out a bit?

Definitely a hard balance to strike between dev-portal & marketing properties.

In other components, I think we rely on _not_ setting specific font stacks inline, and instead leaving `font-family` inherited, or using `var(--font-body)` which we have a bit more control over.

```suggestion
  font-family: var(--font-body);
```

zchsh (commented):
🙌 This looks really close to me, possibly already there?

My only concern is the explicit system stack for `font-family`. Seems fine for `dev-portal`, but seems like it'll be strange to have this be one instance on dot-io and marketing sites of the system font stack. Since this component is shared across properties which don't all use the system font stack, I think we should find a solution to make the font family flexible - I think `var(--font-body)` works in other components (thanks to your "core" dev-portal layout workarounds, IIRC!)

ashleemboyer:
@zchsh updated these items in https://github.com/hashicorp/react-components/pull/547/commits/7e048f8a214f9775751d56f8d4a134608a3b6e21!

BRKalow (commented):
For publishing, can you run `npx changeset` locally and commit up a changeset file? I think this would be a minor bump.

ashleemboyer:
@BRKalow done! Will the new version automatically be published now on merge?

PR #552, opened 2022-03-28
feat(docs-page): support latestVersionRef option to RemoteContentLoader

BRKalow:
🎟️ [Asana Task]()

---

## Description

Supports a `latestVersionRef` option for `RemoteContentLoader` which forces the "latest" version to fetch from a specific ref.

e.g. if we're viewing `waypointproject.io/docs` and we pass `latestVersionRef: 'dev-portal'`, the loader would attempt to load content from `/api/content/waypoint/doc/dev-portal/docs` instead of whatever the latest version is (maybe `/api/content/waypoint/doc/v0.7.x/docs`).

I also fixed a bug with the recent addition to support passing `remarkPlugins` as a function.

### PR Checklist 🚀

### DIFF
--- a/packages/docs-page/server/loaders/remote-content.ts
+++ b/packages/docs-page/server/loaders/remote-content.ts
@@ -21,6 +21,11 @@ interface RemoteContentLoaderOpts extends DataLoaderOpts {
   remarkPlugins?: ((params?: ParsedUrlQuery) => $TSFixMe[]) | $TSFixMe[]
   mainBranch?: string // = 'main',
   scope?: Record<string, $TSFixMe>
+  /**
+   * Allows us to override the default ref from which we fetch the latest content.
+   * e.g. when no version exists in the path, and latestVersionRef was 'my-stable-branch', we would fetch content for `my-stable-branch`
+   */
+  latestVersionRef?: string
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/docs-page/server/loaders/remote-content.ts
```diff
@@ -125,6 +132,9 @@ export default class RemoteContentLoader implements DataLoader {
           '`remarkPlugins:` When specified as a function, must return an array of remark plugins'
         )
       }
+    } else {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- we default this in the constructor, so it must be defined
+      remarkPlugins = this.opts.remarkPlugins!
```
dstaley, line 137: Not for this PR, but it might be nice refactor `opts` into separate class properties so we don't need to do stuff like this.

packages/docs-page/server/loaders/remote-content.ts
```diff
@@ -186,9 +199,10 @@ export default class RemoteContentLoader implements DataLoader {
     if (document.githubFile) {
       // Link latest version to `main`
       // Hide link on older versions
-      const isLatest = versionMetadataList.find(
-        (e) => e.version === document.version
-      )!.isLatest
+      const isLatest =
+        (versionFromPath === 'latest' && Boolean(this.opts.latestVersionRef)) ||
+        versionMetadataList.find((e) => e.version === document.version)!
```
dstaley, line 204: Is this non-null assertion safe to do (in that `versionMetadataList` will _always_ have a vertion that matches `document.version`)?

dstaley (approved):

packages/docs-page/server/loaders/remote-content.ts
BRKalow: Yeah, if we get to this point that should be the case.

PR #568, opened 2022-04-19
feat: convert packages to TypeScript

dstaley:
🎟️ [Asana Task]()

---

## Description

This PR adds TypeScript types to more packages (mainly those also used in Dev Portal).

The `typecheck` step is expected to fail since we're introducing breaking changes in some downstream packages, and the `typecheck` step uses local versions of packages instead of the version defined in `package.json`.

### PR Checklist 🚀

### DIFF
--- a/packages/code-block/partials/clipboard-button/index.tsx
+++ b/packages/code-block/partials/clipboard-button/index.tsx
@@ -7,11 +7,16 @@ import svgCopySuccess from './svg/copy-success.svg?include'
 import s from './style.module.css'
 import analytics, { heapAttributes } from '../../analytics'
 
-function ClipboardButton({ className, getText }) {
+interface ClipboardButtonProps {
+  className?: string
+  getText: () => Promise<[unknown, null] | [null, string]>
+}
+
+function ClipboardButton({ className, getText }: ClipboardButtonProps) {
   // copiedState can be null (initial), true (success), or false (failure)
-  const [copiedState, setCopiedState] = useState()
+  const [copiedState, setCopiedState] = useState<boolean | null>(null)
   // we reset copiedState to its initial value using a timeout
-  const [resetTimeout, setResetTimeout] = useState()
+  const [resetTimeout, setResetTimeout] = useState<number>()
 
   // Handle copy button clicks
   async function onClick() {
@@ -20,7 +25,8 @@ function ClipboardButton({ className, getText }) {
     // If text cannot be retrieved, exit early to handle the error
     if (getTextError) return handleError(getTextError)
     // Otherwise, continue on...
-    const isCopied = await copyToClipboard(text)
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const isCopied = copyToClipboard(text!)
     // If there's an internal failure copying text, exit early to handle the error
     if (!isCopied) return handleError(`ClipboardButton failed. Text: ${text}.`)
     // Otherwise, things went well, track the event and set state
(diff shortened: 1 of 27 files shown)


packages/select-input/index.tsx
```diff
@@ -3,6 +3,16 @@ import classNames from 'classnames'
 import Downshift from 'downshift'
 import s from './style.module.css'
 
+interface SelectInputProps {
+  name: string
```
dstaley, line 7: I'm actually not completely sure this should be a required prop. Normally I'd say all `input` elements need one, but since this is being used in a `type="hidden"` I think it's technically safe to omit it.

dstaley:
These package updates are all passing typechecking in Dev Portal :tada:

packages/code-block/partials/clipboard-button/index.tsx
```diff
@@ -41,14 +47,14 @@ function ClipboardButton({ className, getText }) {
...
-    clearTimeout(resetTimeout)
+    window.clearTimeout(resetTimeout)
     // Only run the copiedState reset if it's needed
     const needsReset = copiedState != null
     if (needsReset) {
       // Let failure messages linger a bit longer
       const resetDelay = copiedState == false ? 4000 : 1750
       // Set the timeout to reset the copy success state
-      setResetTimeout(setTimeout(() => setCopiedState(null), resetDelay))
+      setResetTimeout(window.setTimeout(() => setCopiedState(null), resetDelay))
```
BRKalow, line 57: Curious, why the need to use the `window.` prefix here?

packages/code-block/partials/code-block-config/index.tsx
```diff
@@ -1,7 +1,14 @@
 import React, { Children } from 'react'
-import CodeBlock from '../../index.js'
+import type { ReactElement } from 'react'
+import CodeBlock from '../../index'
+import type { CodeBlockProps, CodeBlockOptions } from '../../index'
 import normalizePlainCode from '../../utils/normalize-plain-code'
 
+export interface CodeBlockConfigProps extends CodeBlockProps, CodeBlockOptions {
+  children?: ReactElement[]
+  hideClipboard?: boolean
```
BRKalow, line 9: showClipboard and hideClipboard 😄

packages/code-block/partials/code-tabs/index.tsx
```diff
@@ -38,7 +56,9 @@ function CodeTabs({ children, heading, className, tabs, theme = 'dark' }) {
       ]
       const matchIdx = validComponents
         .map((c) => c.component)
-        .indexOf(tabChild.type)
+        // we use `never` here instead of `any` to signify that we don't care
+        // about the type
```
BRKalow, line 60: 👍

packages/code-block/provider/index.tsx
```diff
@@ -24,7 +32,7 @@ export default function CodeTabsProvider({ children }) {
       // the preference array
       .concat(preferredTabGroups)
       // Remove any falsy values (eg initial)
-      .filter((s) => s !== null && s !== '')
+      .filter((s): s is string => s !== null && s !== '')
```
BRKalow, line 35: neat 👍

packages/text-split/index.tsx
```diff
@@ -32,7 +52,7 @@ function TextSplit({
             </h2>
           )}
           {hasStringContent ? (
-            <ContentString contentString={content} theme={theme} />
+            <ContentString contentString={content} />
```
BRKalow, line 55: Why'd we remove the theme prop here?

BRKalow (approved):
Looks good overall, doesn't seem like you meaningfully changed anything! Nice work.

packages/code-block/partials/clipboard-button/index.tsx
dstaley: It's a node/dom thing in TypeScript. At least how we have it configured, `setTimeout` returns a `NodeJS.Timeout`, whereas `clearTimeout` accepts a `number`. By using the `window.` prefix, we are able to get a reference to the dom versions of those APIs. Since this is being called in a `useEffect`, it should be safe to rely on `window`.

packages/text-split/index.tsx
dstaley: Because `ContentString` only uses the `contentString` prop. https://github.com/hashicorp/react-components/blob/main/packages/text-split/index.js#L53

PR #583, opened 2022-05-05
feat(RemoteContentLoader): navDataPrefix option

thiskevinwang:
# Description

This adds a `navDataPrefix` option to our `RemoteContentLoader`.

This helps to address a previous assumption that `basePath` would always be the same for `doc` and `nav-data` lookups from our API, which is not necessarily going to be the case going forward.

🎟️ [Asana Task]()

---

## Description

### PR Checklist 🚀

### DIFF
--- a/packages/docs-page/server/loaders/remote-content.ts
+++ b/packages/docs-page/server/loaders/remote-content.ts
@@ -17,6 +17,13 @@ import { getPathsFromNavData } from '../get-paths-from-nav-data'
 
 interface RemoteContentLoaderOpts extends DataLoaderOpts {
   basePath: string
+  /**
+   * In most cases, `basePath` should suffice when resolving nav-data, because
+   * it happens to match the prefix of nav-data file.
+   *
+   * If it does not, `navDataPrefix` will serve as an optional override.
+   */
+  navDataPrefix?: string
   enabledVersionedDocs?: boolean
   remarkPlugins?: ((params?: ParsedUrlQuery) => $TSFixMe[]) | $TSFixMe[]
   mainBranch?: string // = 'main',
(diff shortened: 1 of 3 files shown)


BRKalow:
Ah I see now. So generally it's `{basePath}-nav-data.json`, but we need a way to change that. :nod:

packages/docs-page/server/loaders/remote-content.ts
```diff
@@ -95,22 +104,28 @@ export default class RemoteContentLoader implements DataLoader {
     if (!this.opts.mainBranch) this.opts.mainBranch = 'main'
     if (!this.opts.scope) this.opts.scope = {}
     if (!this.opts.remarkPlugins) this.opts.remarkPlugins = []
+
+    this.navDataPrefix = this.opts.navDataPrefix || this.opts.basePath
```
BRKalow, line 108: suggestion: convention is for everything to be read from `this.opts`, so we can use a similar default approach above and assign to `this.opts.basePath` if it is unset. No need for the extra class property

thiskevinwang:
> Ah I see now. So generally it's `{basePath}-nav-data.json`, but we need a way to change that. :nod:

Yea, this has been _one of those assumptions_ that hasn't caused issues in the past, but not _sort of_ does

packages/docs-page/server/loaders/remote-content.ts
thiskevinwang: Good call! I'll do that

thiskevinwang: https://github.com/hashicorp/react-components/pull/583/commits/445d4fbef12bddd77ee2af5997b8f94518e765e0

BRKalow (approved):
LGTM!

PR #591, opened 2022-05-18
feat(code-block): onCopyCallback

zchsh:
🎟️ [Asana Task](https://app.asana.com/0/1201903760348480/1202114097062999/f)
🔍 [Preview Link](https://react-components-git-zscode-block-on-copy-hashicorp.vercel.app/components/codeblock)

---

This PR implements an `onCopyCallback` prop for `code-block`. This callback is called after `code-block`'s Copy button is clicked.

Note that this callback is _not_ called as a direct result of the `onClick` on the button - instead, we wait until we can confirm whether the copying of text was successful.

### PR Checklist 🚀

### DIFF
--- a/packages/code-block/index.test.js
+++ b/packages/code-block/index.test.js
@@ -6,6 +6,7 @@ import {
   waitFor,
 } from '@testing-library/react'
 import CodeBlock from './'
+import { heapAttributes } from './analytics'
 // Mock copy-to-clipboard
 import copyToClipboard from './partials/clipboard-button/copy-to-clipboard'
 jest.mock('./partials/clipboard-button/copy-to-clipboard', () =>
@@ -76,6 +77,7 @@ it('should have a data-heap-track="code-block-clipboard-icon" attribute on the C
   )
   const copyButton = screen.getByText('Copy')
   expect(copyButton.tagName).toBe('BUTTON')
+  expect(copyButton.getAttribute('data-heap-track')).toBe(heapAttributes.copy)
 })
 
 it('should render a keyboard-focusable `Copy` button', () => {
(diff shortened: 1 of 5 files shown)


packages/code-block/partials/clipboard-button/index.tsx
```diff
@@ -45,6 +50,10 @@ function ClipboardButton({ className, getText }: ClipboardButtonProps) {
   // reset to the default appearance so that it's clear
   // the "Copy" button can be used again
   useEffect(() => {
```
zchsh, line 52: We wait till the `copiedState` has changed before calling the callback.

This way, the callback has richer information, as it provides a `copiedState`, rather than only reporting that the copy button was clicked (with an unknown outcome) as we'd get if we hooked into the copy button's `onClick`.

packages/code-block/index.test.js
```diff
@@ -76,6 +77,7 @@ it('should have a data-heap-track="code-block-clipboard-icon" attribute on the C
   )
   const copyButton = screen.getByText('Copy')
   expect(copyButton.tagName).toBe('BUTTON')
+  expect(copyButton.getAttribute('data-heap-track')).toBe(heapAttributes.copy)
```
zchsh, line 80: This and the `heapAttributes` import are an unrelated fix for a slip I noticed when writing new tests. Previously we weren't actually testing for the `data-heap-track` attribute 😅

zchsh:
@BrandonRomano This change is mainly targeted at enabling https://github.com/hashicorp/dev-portal/pull/443. Let me know what you think 🙇‍♂️

BrandonRomano (approved):
Nice improvement here ✨ 

Was there an observed problem with the simple click handling, or just hardening to ensure it was actually copied?

PR #603, opened 2022-06-06
fix(docs-page): allow trailing index pages to 404

zchsh:
- fix(docs-page): remove index path part resolution in remote loader
- chore(docs-page): add changeset

🎟️ [Asana Task]()

---

## Description

### PR Checklist 🚀

### DIFF
--- a/packages/docs-page/server/loaders/remote-content.ts
+++ b/packages/docs-page/server/loaders/remote-content.ts
@@ -162,10 +162,6 @@ export default class RemoteContentLoader implements DataLoader {
 
     const versionMetadataList: VersionMetadataItem[] =
       await cachedFetchVersionMetadataList(this.opts.product)
-    // remove trailing index to ensure we fetch the right document from the DB
-    const pathParamsNoIndex = paramsNoVersion.filter(
-      (param, idx, arr) => !(param === 'index' && idx === arr.length - 1)
-    )
 
     const latestVersion =
       this.opts.latestVersionRef ??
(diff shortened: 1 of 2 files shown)


packages/docs-page/server/loaders/remote-content.ts
```diff
@@ -162,10 +162,6 @@ export default class RemoteContentLoader implements DataLoader {
 
     const versionMetadataList: VersionMetadataItem[] =
       await cachedFetchVersionMetadataList(this.opts.product)
-    // remove trailing index to ensure we fetch the right document from the DB
-    const pathParamsNoIndex = paramsNoVersion.filter(
-      (param, idx, arr) => !(param === 'index' && idx === arr.length - 1)
-    )
```
zchsh, line 168: I _think_ removing this gets us the desired behaviour. But, I feel like I might be missing additional context here. My assumption here:
- The only situation in which incoming URL `params` parts would have a trailing `index`, AND we want to resolve and render the page at that URL, is for pages that are genuinely called `index`.
    - And such pages should resolve from the content API _even with_ the trailing index in their `fullPath`

zchsh:
@BRKalow I think the redirect rule is great! Big +1 to looking for increased 404s as well.

Any reason not to do both the redirect rule, and ensure content doesn't get resolved when it's not supposed to?

BRKalow:
I think if we don't need to add the redirect rule then we shouldn't do so. Agreed that if content isn't meant to exist at a given URL we definitely shouldn't be returning a successful response! I didn't mean to imply that your change wasn't necessary, I think this is definitely something we want to hedge against.

BRKalow (approved):
I found the original PR where I introduced it, but unfortunately I didn't leave any context as to _why_ 😅 https://github.com/hashicorp/react-components/pull/233

I guess we'll just want to keep an eye on our analytics and ensure there isn't an increase in 404s after this. Nice job!

zchsh:
@BRKalow  Ah cool okay, apologies if that came off the wrong way 💖  👯 does this PR look good to you? I've proven it out in https://github.com/hashicorp/dev-portal/pull/556 and it _seems_ to be working.

PR #604, opened 2022-06-06
[code-block] Moving invocation of onCopyCallback

ashleemboyer:
🔍 [Preview Link](https://react-components-git-ambadjust-on-copy-callback-hashicorp.vercel.app)

---

## Description

The new `onCopyCallback` function in `CodeBlock` was being invoked on initial load. This PR moves the invocation of this callback to the `onClick` handler for the rendered `<button>` used for copying.

- I've also added some empty lines between blocks of the `onClick` handler and added curly braces to one-line if statements. Both of these sets of additional changes were made to improve the readability of this code.

## Test steps

A draft PR was created to test these changes in `dev-portal`. See [dev-portal#555](https://github.com/hashicorp/dev-portal/pull/555).

- [ ] Go to https://dev-portal-git-ambtest-on-copy-callback-hashicorp.vercel.app/waypoint/downloads
- [ ] No toast should appear on load
- [ ] Click the Copy button on the CodeBlock shown
- [ ] A green success toast should appear

### PR Checklist 🚀

### DIFF
--- a/.changeset/clever-zebras-shout.md
+++ b/.changeset/clever-zebras-shout.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-code-block': patch
+---
+
+Moves the invocation of onCopyCallback to after the Copy button has been clicked.
--- a/.changeset/clever-zebras-shout.md
+++ b/.changeset/clever-zebras-shout.md
@@ -1,5 +1,0 @@
----
-'@hashicorp/react-code-block': patch
----
-
-Moves the invocation of onCopyCallback to after the Copy button has been clicked.
--- a/.changeset/wise-terms-heal.md
+++ b/.changeset/wise-terms-heal.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-code-block': major
+---
+
+Moves the invocation of `onCopyCallback` to after the copy button has been clicked, and only invokes the callback if `copiedState` is `true` or `false`.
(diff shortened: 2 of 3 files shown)


zchsh:
@ashleemboyer I think this might be a breaking change - the intent with the previous implementation from #591:

- On load, and also on reset of copied state after timeout, `onCopyCallback(null)` is called
- On successful copy, `onCopyCallback(true)` is called
- On failed copy, `onCopyCallback(false)` is called
Intent being to provide a bit more information to the consumer.

Hopefully that helps the use case we have! I might be missing why we'd want to change to only call `onCopyCallback` on success.

ashleemboyer:
> @ashleemboyer I think this might be a breaking change - the intent with the previous implementation from #591:
> 
> * On load, and also on reset of copied state after timeout, `onCopyCallback(null)` is called
> * On successful copy, `onCopyCallback(true)` is called
> * On failed copy, `onCopyCallback(false)` is called
> 
> Intent being to provide a bit more information to the component consumer when using this callback.
> 
> Hopefully that helps the use case we have! I might be missing why we'd want to change to only call `onCopyCallback` on success.

@zchsh ahhh, I forgot `onCopyCallback(false)` was valid! I'll update to include that call as well. Thank you for catching that! 😊

zchsh (approved):
Looks good to me! 👍

PR #605, opened 2022-06-06
[code-block] Refocus copy button

ashleemboyer:

---

## Description

### PR Checklist 🚀


ashleemboyer:
Waiting to open for review until #604 is merged.

packages/code-block/partials/clipboard-button/index.tsx
```diff
@@ -50,10 +67,24 @@ function ClipboardButton({
...
+     */
+    if (
+      buttonRef &&
+      buttonRef.current &&
+      document.activeElement === document.body &&
+      document.activeElement !== buttonRef.current
+    ) {
+      buttonRef.current.focus()
     }
+
```
ashleemboyer, line 87: 🗒️ This is the main change for this PR. Other changes will come from #604.

zchsh (approved):

.changeset/wise-terms-heal.md
```diff
@@ -1,5 +0,0 @@
----
-'@hashicorp/react-code-block': major
----
-
-Moves the invocation of `onCopyCallback` to after the copy button has been clicked, and only invokes the callback if `copiedState` is `true` or `false`.
```
ashleemboyer, line 5: Not sure why this is in the diff? 🤔

zchsh (commented):
This looks good! I'm seeing the button retain focus on copy 👍

I'm also noticing when I first load https://react-components-git-ambrefocus-copy-button-hashicorp.vercel.app/components/codeblock, the code block copy button gets focused.

I wonder if we need different logic here to prevent the initial focus grab? Eg maybe `copiedState` should have an initial value of `undefined`, and we can only refocus if `copiedState` is null, true, or false.

ashleemboyer:
> This looks good! I'm seeing the button retain focus on copy 👍
> 
> I'm also noticing when I first load https://react-components-git-ambrefocus-copy-button-hashicorp.vercel.app/components/codeblock, the code block copy button gets focused.
> 
> I wonder if we need different logic here to prevent the initial focus grab? Eg maybe `copiedState` should have an initial value of `undefined`, and we can only refocus if `copiedState` is null, true, or false.

@zchsh good catch! The button shouldn't be focusing on first load. I'll update now!

packages/code-block/partials/clipboard-button/index.tsx
```diff
@@ -73,9 +77,29 @@ function ClipboardButton({
...
+     */
+    if (
+      (copiedState === true || copiedState === false) &&
+      buttonRef &&
+      buttonRef.current &&
+      document.activeElement === document.body &&
+      document.activeElement !== buttonRef.current
+    ) {
+      buttonRef.current.focus()
+    }
```
BRKalow, line 97: Do we know why the button is losing focus? Would it work to attempt to re-focus in our `onClick` handler instead? If it's happening as a side-effect of the copy functionality, it seems like that would be a better spot.

ashleemboyer:
> > This looks good! I'm seeing the button retain focus on copy 👍
> > I'm also noticing when I first load https://react-components-git-ambrefocus-copy-button-hashicorp.vercel.app/components/codeblock, the code block copy button gets focused.
> > I wonder if we need different logic here to prevent the initial focus grab? Eg maybe `copiedState` should have an initial value of `undefined`, and we can only refocus if `copiedState` is null, true, or false.
> 
> @zchsh good catch! The button shouldn't be focusing on first load. I'll update now!

@zchsh updated in https://github.com/hashicorp/react-components/pull/605/commits/c75f3f5402ea7b294f3aeaf2f18b9f844b1730eb!

packages/code-block/partials/clipboard-button/index.tsx
ashleemboyer: Yes, the button is losing focus because it's getting re-rendered when `setCopiedState` is called. Let me see if calling `.focus` is possible right after `setCopiedState(true)` and `setCopiedState(false)`.

ashleemboyer: @BRKalow I added an `updateCopiedState` helper in https://github.com/hashicorp/react-components/pull/605/commits/866849cec9cb466ab12e5e7c6c6304802d953084. How does it look?

```tsx
/**
 * Handle setting state, re-focusing the button, and invoking the copy
 * callback. This should not be called when `copiedState` is being reset, only
 * when there is a success or failure of the copy action.
 */
function updateCopiedState(newState: boolean) {
  // Update the local state variable
  setCopiedState(newState)

  // Re-focus the copy button
  buttonRef?.current?.focus()

  // If an onCopyCallback was provided
  if (typeof onCopyCallback == 'function') {
    onCopyCallback(newState)
  }
}
```

zchsh (approved):

packages/code-block/partials/clipboard-button/index.tsx
```diff
@@ -46,26 +50,34 @@ function ClipboardButton({
...
+   * Handle setting state, re-focusing the button, and invoking the copy
+   * callback. This should not be called when `copiedState` is being reset, only
+   * when there is a success or failure of the copy action.
+   */
+  function updateCopiedState(newState: boolean) {
+    // Update the local state variable
+    setCopiedState(newState)
+
+    // Re-focus the copy button
+    buttonRef?.current?.focus()
```
BRKalow, line 76:
```suggestion
    // Re-focus the copy button as it loses focus when we call `.select()` on the textarea element in `copyToClipboard`
    buttonRef?.current?.focus()
```

BRKalow (approved):
Looks good, minor clarification suggestion on a comment!

packages/code-block/partials/clipboard-button/index.tsx
ashleemboyer: Updated in https://github.com/hashicorp/react-components/pull/605/commits/9490b914e6daf6d23fee60979bed4b5e850b9f58! Also updated the `onCopyCallback` description in https://github.com/hashicorp/react-components/pull/605/commits/54863c94f759d7ce1dcf123d0379fef74bb2c17f, which I should've done in the last PR. 😁

PR #607, opened 2022-06-07
feat: update docs-page utils for anchor links

zchsh:
🎟️ [Asana Task](https://app.asana.com/0/1202114367927919/1202384957391166/f)
🔍 [Preview Link](https://react-components-git-zsbump-docs-page-utils-hashicorp.vercel.app)

---

This PR updates `@hashicorp/react-docs-page` to drop the `@hashicorp/platform-markdown-utils` dependency, by exposing `remarkPlugins` and `rehypePlugins` options directly to the consumer, rather than inheriting a fixed set of defaults.

This change is being made so that in `dev-portal`, we can run exactly the set of remark and rehype plugins we want, rather than having to work around the same fixed set of defaults for both `dot-io` and `dev-dot` docs views.

### PR Checklist 🚀


zchsh:
Note: squeaked in some very minor CSS fixes for scroll offset for "inline code list items" here. Have tested in https://github.com/hashicorp/dev-portal/pull/559 👍

PR #614, opened 2022-06-15
Rename event to error page loaded

BrandonRomano:
## Description

Update our error page analytics event to be consistently named.

This will make Analysis against this event easier, as the way that the event is currently structured produces a new table in our analysis systems, which makes broad error analysis nearly impossible to do.

We're only sending along the error code itself, as Segment will automatically send two context values which will be useful for error analysis: `context_page_url` (the URL where the error happened) and `context_page_referrer` (the URL before the error happened, which will be useful for sniffing out the source of 404s).

### PR Checklist 🚀

### DIFF
--- a/.changeset/nervous-gorillas-hear.md
+++ b/.changeset/nervous-gorillas-hear.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-error-view': patch
+---
+
+Adjust error analytics event to have a consistent name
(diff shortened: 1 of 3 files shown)


BRKalow (approved):

BRKalow:
@BrandonRomano can you add a changeset file? Follow the directions in the comment above

BRKalow:
And looks like there's a test failing 👀

hashicorpzed (commented):
:chefs_kiss:

PR #618, opened 2022-06-16
feat: `<MotionConfig />` component

alexcarpenter:
🎟️ [Asana Task]()

---

## Description

Introduce a `MotionConfig` component that handles lazing loading Framer Motion features. Ref [reduce bundle size](https://www.framer.com/docs/guide-reduce-bundle-size/#available-features).

### Usage

```tsx
// _app.tsx
import MotionConfig from '@hashicorp/react-motion-config'

export default function App({ Component, pageProps }) {
  return (
    <MotionConfig>
      <Component {...pageProps} />
    </MotionConfig>
  )
}
```

### PR Checklist 🚀

### DIFF
--- a/packages/motion-config/package.json
+++ b/packages/motion-config/package.json
@@ -1,0 +1,13 @@
+{
+  "name": "@hashicorp/react-motion-config",
+  "version": "0.1.0",
+  "description": "",
+  "author": "HashiCorp",
+  "license": "MPL-2.0",
+  "dependencies": {
+    "framer-motion": "^6.3.11"
+  },
+  "peerDependencies": {
+    "react": ">=16.x"
+  }
+}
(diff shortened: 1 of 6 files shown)


packages/motion-config/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@hashicorp/react-motion-config",
+  "version": "0.1.0",
+  "description": "",
+  "author": "HashiCorp",
+  "license": "MPL-2.0",
+  "dependencies": {
+    "framer-motion": "^6.3.11"
```
BRKalow, line 8: I wonder if we want to make this a peer dep? 🤔 Since there is a chance we could end up with two versions depending on what version the consumer has installed

alexcarpenter: That is a good callout, I would assume so based on my limited understanding and what I am reading.

BRKalow (approved):
neat 👍 changeset too?

PR #623, opened 2022-06-20
feat(hashi-head): validate props.image, add twitter:description

zchsh:
🎟️ [Asana Task](https://app.asana.com/0/1100423001970639/1202415308965256/f)
🔍 [Preview Link](https://react-components-git-zsmeta-tags-for-social-share-cards-hashicorp.vercel.app)

---

This PR adds validation to try to ensure that the `image` prop provided to `@hashicorp/react-head` is an absolute URL. This is to match the Open Graph protocol, which notes [All valid URLs that utilize the https:// or https:// protocols](https://ogp.me/#url).

### PR Checklist 🚀

### DIFF
--- a/packages/head/helpers/is-absolute-url.ts
+++ b/packages/head/helpers/is-absolute-url.ts
@@ -1,0 +1,17 @@
+/**
+ * Given a string,
+ * return true if the string is an absolute URL,
+ * or false otherwise.
+ *
+ * Uses regex so may not be 100% accurate.
+ * Based on https://github.com/sindresorhus/is-absolute-url
+ *
+ * @param string The URL to test
+ * @returns
+ */
+function isAbsoluteUrl(string: string): boolean {
+  const regex = /^[a-zA-Z][a-zA-Z\d+\-.]*:/
+  return regex.test(string)
+}
+
+export default isAbsoluteUrl
--- a/packages/head/helpers/is-absolute-url.ts
+++ b/packages/head/helpers/is-absolute-url.ts
@@ -7,7 +7,7 @@
  * Based on https://github.com/sindresorhus/is-absolute-url
  *
  * @param string The URL to test
- * @returns
+ * @returns true if the URL is absolute, false otherwise
  */
 function isAbsoluteUrl(string: string): boolean {
   const regex = /^[a-zA-Z][a-zA-Z\d+\-.]*:/
--- a/packages/head/helpers/is-absolute-url.ts
+++ b/packages/head/helpers/is-absolute-url.ts
@@ -6,6 +6,15 @@
  * Uses regex so may not be 100% accurate.
  * Based on https://github.com/sindresorhus/is-absolute-url
  *
+ * @example
+ * // returns true, this is an absolute URL
+ * isAbsoluteUrl('https://www.hashicorp.com/foo/bar');
+ * @example
+ * // returns false, this is a relative path
+ * isAbsoluteUrl('./foo/bar');
+ * @example
+ * // returns false, this is an absolute path, but not an absolute URL
+ * isAbsoluteUrl('/foo/bar');
  * @param string The URL to test
  * @returns true if the URL is absolute, false otherwise
  */
--- a/packages/head/index.tsx
+++ b/packages/head/index.tsx
@@ -10,7 +10,7 @@ export default function HashiHead(props: HashiHeadProps): React.ReactElement {
   /**
    * Throw an error if props.image is a relative URL.
    * It must be an absolute URL in order to work as expected as og:image.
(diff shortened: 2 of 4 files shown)


packages/head/index.tsx
```diff
@@ -24,12 +49,20 @@ export default function HashiHead(props: HashiHeadProps): React.ReactElement {
...
-        />
+        <>
+          <meta
+            name="description"
+            property="og:description"
+            content={props.description}
+            key="description"
+          />
+          <meta
+            // TODO: are we sure we need this? Seems like will fall back to OG
```
zchsh, line 60: @BRKalow wanted to confirm - do we strictly need `twitter:description`? It seems like it should fall back to `og:description` (ref: https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/markup). Wondering whether we should omit it (or perhaps duplicate all such fallback tags with `twitter:` specific versions, for consistency?)

packages/head/helpers/is-absolute-url.ts
```diff
@@ -0,0 +1,17 @@
...
+ * return true if the string is an absolute URL,
+ * or false otherwise.
+ *
+ * Uses regex so may not be 100% accurate.
+ * Based on https://github.com/sindresorhus/is-absolute-url
+ *
+ * @param string The URL to test
+ * @returns true if the URL is absolute, false otherwise
+ */
+function isAbsoluteUrl(string: string): boolean {
```
zchsh, line 12: Pulled in from hashicorp/dev-portal (and elsewhere).

packages/head/index.tsx
```diff
@@ -1,9 +1,34 @@
...
+
 export default function HashiHead(props: HashiHeadProps): React.ReactElement {
+  /**
+   * Throw an error if props.image is a relative URL.
+   * It must be an absolute URL in order to work as expected as og:image.
+   * Reference: https://ogp.me/#url
+   */
+  if (typeof props.image !== 'undefined' && !isAbsoluteUrl(props.image)) {
+    /**
+     * TODO: should we consider alternatives to throwing an error here?
```
zchsh, line 17: @BRKalow Forget to mention in review - wanted to get your take on the error approach here. I think the `IS_DEV` approach could make sense, but would love a second opinion on that (and also a lil curious about whether further tracking, eg in DataDog if available, sounds like something we might want to do).

packages/head/helpers/is-absolute-url.ts
ashleemboyer: (Question 🙋) Can some examples of what is and isn't an absolute URL be added to the block comment above this function? That will help make it clear at a glance how this function works. I end up having to look it up every time. 😅

(Future 📌) Tests would be nice as well, but maybe this is a helper we should prioritize moving to [`@hashicorp/platform-util`](https://github.com/hashicorp/web-platform-packages/tree/main/packages/util) first? Sounds like it's very widely used!

packages/head/index.tsx
ashleemboyer: I think you should be able to test this both with & without `twitter:description` using the [Twitter card validator](https://cards-dev.twitter.com/validator). Have you tried it out yet to help confirm the necessity of this tag?

packages/head/index.tsx
ashleemboyer: If our goal is to fix all existing now-incorrect values for `props.image,` it seems reasonable to me to always throw an error regardless of the `NODE_ENV`. No production builds will be immediately broken unless the version for this package is upgraded after release right? We'll just start seeing failed Vercel builds in PRs?

zchsh: ☝️ I think we'd definitely want to always throw an error if we did 100% static builds, but might not be as safe anymore since we don't static render all pages... for ISR pages, I think throwing an error could cause a client-side error, or possibly a 500 for the initial server-side render? Does that sound accurate?

packages/head/index.tsx
ashleemboyer: UGH that's so annoying 😭 but great tests! 🙌🏻 Glad that was fairly straightforward to test.

packages/head/helpers/is-absolute-url.ts
ashleemboyer: These examples look great! Very clear and super helpful. Thank you for adding 💖

packages/head/index.tsx
ashleemboyer: Hmm, yeah. That makes sense. I wonder if we can test different cases to make sure? But that can definitely be done outside of this PR! We can always update this block in the future if needed. 📌

ashleemboyer (approved):
This looks good to me! ✨

PR #630, opened 2022-06-29
Add `marketo-form` component

dstaley:
🎟️ [Asana Task]()

---

## Description

### PR Checklist 🚀

### DIFF
--- a/packages/marketo-form/server/api-routes.ts
+++ b/packages/marketo-form/server/api-routes.ts
@@ -1,0 +1,71 @@
+import { URL } from 'url'
+import createFetch from '@vercel/fetch'
+import type { NextApiRequest, NextApiResponse } from 'next'
+
+const fetch = createFetch()
+
+interface MarketoTokenResponse {
+  access_token: string
+}
+
+async function getToken(): Promise<MarketoTokenResponse> {
+  const url = new URL(`${process.env.MARKETO_IDENTITY}/oauth/token`)
+  url.searchParams.append('grant_type', 'client_credentials')
+  url.searchParams.append('client_id', process.env.MARKETO_CLIENT_ID!)
+  url.searchParams.append('client_secret', process.env.MARKETO_CLIENT_SECRET!)
+  const res = await fetch(url.toString())
+  return await res.json()
+}
+
+async function getForm(req: NextApiRequest, res: NextApiResponse) {
+  try {
+    const { access_token } = await getToken()
+    const marketoRes = await fetch(
+      `${process.env.MARKETO_ENDPOINT}/asset/v1/form/${req.query.form}/fields.json`,
+      {
+        headers: {
+          'Content-Type': 'application/json',
+          Authorization: `Bearer ${access_token}`,
+        },
+      }
+    )
+    const form = await marketoRes.json()
+    return res.status(marketoRes.status).json(form)
+  } catch (err) {
+    console.error(err)
+    return res.status(500)
+  }
+}
+
+async function submitForm(req: NextApiRequest, res: NextApiResponse) {
+  try {
+    const { access_token } = await getToken()
+    const marketoRes = await fetch(
+      `${process.env.MARKETO_ENDPOINT}/v1/leads/submitForm.json`,
+      {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          Authorization: `Bearer ${access_token}`,
+        },
+        body: JSON.stringify(req.body),
+      }
+    )
+    const form = await marketoRes.json()
+    return res.status(marketoRes.status).json(form)
+  } catch (err) {
+    console.error(err)
+    return res.status(500)
+  }
+}
+
+export default async (req: NextApiRequest, res: NextApiResponse) => {
+  switch (req.query.marketo[0]) {
+    case 'form':
+      return getForm(req, res)
+    case 'submit':
+      return submitForm(req, res)
+    default:
+      return res.status(404).json({ error: 'not found' })
+  }
+}
--- a/.changeset/odd-lemons-bathe.md
+++ b/.changeset/odd-lemons-bathe.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-marketo-form': minor
+---
+
+Initial release
--- a/package-lock.json
+++ b/package-lock.json
@@ -3938,6 +3938,10 @@
       "resolved": "packages/markdown-page",
       "link": true
     },
+    "node_modules/@hashicorp/react-marketo-form": {
+      "resolved": "packages/marketo-form",
+      "link": true
+    },
     "node_modules/@hashicorp/react-min-100-layout": {
       "resolved": "packages/min-100-layout",
       "link": true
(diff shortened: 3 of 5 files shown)


packages/alert-banner/index.tsx
```diff
@@ -60,7 +60,7 @@ function AlertBanner({
    * to remember the dismissal of this banner
    */
   function closeBanner() {
-    cookie.set(dismissalCookieId, 1)
+    cookie.set(dismissalCookieId, '1')
```
BRKalow, line 63: 🤔  was this a type error?

packages/marketo-form/form/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+          disabled={methods.formState.isSubmitting}
+          title={
+            methods.formState.isSubmitting
+              ? 'Submitting...'
+              : submitTitle ?? 'Submit'
+          }
+        />
+        <div>
+          <small>{`Marketo Form (${formId})`}</small>
+        </div>
```
BRKalow, line 130: I'm guessing this is for testing? 💡 It could be nice to have this only render in dev, if it's something that would be helpful long-term.

packages/marketo-form/form/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+interface Props {
+  formId: number
+  marketoForm: MarketoForm
+  groups?: Record<
+    string,
+    {
+      fields: string[]
+      component: (props: { fields: MarketoFormField[] }) => JSX.Element
+    }
```
BRKalow, line 25: nit: It could be nice to pull the Group type out into a separate interface.

.github/workflows/canary-release.yml
```diff
@@ -15,7 +15,9 @@ jobs:
     if: contains( github.event.pull_request.labels.*.name, 'release:canary')
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@master
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
BRKalow, line 20: This addresses the warning we're seeing, right?

packages/marketo-form/partials/fields/checkbox-field/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+import Wrapper from '../../wrapper'
+import { formattedLabel } from '../../../utils'
+import type { MarketoFormCheckboxField } from '../../../types'
+
+const CheckboxField = ({ field }: { field: MarketoFormCheckboxField }) => {
+  const { register, watch } = useFormContext()
+  const { errors, touchedFields } = useFormState()
+  const value = watch(field.id, false)
+
+  if (field.id === 'Consent_Privacy_Policy__c') {
```
BRKalow, line 12: 💭 : This feels like a concern that shouldn't live in this component. I know you called it out in your earlier video so it's on your mind, but just noting that here. I don't think we need to address it immediately, but something to keep tabs on.

packages/marketo-form/partials/fields/select-field/index.tsx
```diff
@@ -0,0 +1,62 @@
+import { useMemo } from 'react'
+import { useFormContext } from 'react-hook-form'
+import SelectInput from '@hashicorp/react-select-input'
+import Wrapper from '../../wrapper'
+import type { MarketoFormSelectField, SelectValue } from '../../../types'
+
+function applyVisibilityRule(
```
BRKalow, line 7: Just a gut check, the visibility rules only apply to select fields right now? Is this feature available for other types of fields?

packages/marketo-form/partials/wrapper/index.tsx
```diff
@@ -0,0 +1,17 @@
+import type { ReactNode } from 'react'
+import clsx from 'clsx'
+import styles from './style.module.css'
+
+const Wrapper = ({ size, children }: { size?: 'lg'; children: ReactNode }) => {
```
BRKalow, line 5: nit: `FieldWrapper` might help provide a little more context when seeing usage.

packages/marketo-form/form/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+    return groupFields(groups, marketoForm.result)
+  }, [marketoForm, groups])
+
+  const methods = useForm({
+    mode: 'onBlur',
+    defaultValues: calculateDefaultValues(marketoForm.result),
+  })
+
+  const onSubmit = async (data: Record<string, unknown>) => {
+    const res = await fetch(`${window.location.origin}/api/marketo/submit`, {
```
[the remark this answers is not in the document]
dstaley: Yup, as you caught below this is to account for the fact that in test environments `fetch` doesn't resolve hostname-less URLs.

packages/alert-banner/index.tsx
dstaley: Yup, I added the [types package](https://github.com/hashicorp/react-components/pull/630/files#diff-f7917ea701516a60f29f5b1677192ac32f5deaa9f624ee55e6fc752e460275b5R30) to (eventually) support UTM params from cookies, and caught this. I'll probably move this out to a separate PR though.

packages/marketo-form/form/index.tsx
dstaley: Yup. I'll move this to a `data` attribute on the form moving forward so that we can still easily confirm the destination Marketo form even in production.

packages/marketo-form/partials/fields/checkbox-field/index.tsx
dstaley: Yeah, so there's three ways to add this functionality: in the base form field component as shown here, in the disambiguation component (as demonstrated [here](https://github.com/hashicorp/react-components/pull/630/files#diff-34895b2691e11077be55bd2655a9652901f961bcc6ce6c653a649b95c66605a1R34)), and finally as a default group, for example:

```ts
const defaultFieldGroupings = {
  name: {
    fields: ['FirstName', 'LastName'],
    component: NameField,
  },
  privacy: {
    fields: ['Consent_Privacy_Policy__c'],
    component: PrivacyCheckbox
  }
}
```

I'd prefer one of the first two over the third, since it mixes concepts of default field controls with default field groupings. However I agree with you that it doesn't feel quite right embedding that logic into the base form control component. The second approach (that the country combo box is using) feels slightly better, but it also still doesn't feel like the _right_ place.

packages/marketo-form/partials/fields/select-field/index.tsx
dstaley: Correct, I discovered recently that we don't use visibility rules in any of our forms, so I'll likely be removing this code in a future iteration. Marketo supports applying visibility rules to any field, but given that we aren't using it currently it doesn't make sense to keep this in.

packages/marketo-form/partials/fields/checkbox-field/index.tsx
BRKalow: The disambiguation seems like a reasonable starting point to me! Agreed it doesn't feel fully _right_, but I think it's a reasonable starting point! 👍

packages/marketo-form/server/api-routes.ts
```diff
@@ -92,6 +101,14 @@ async function submitForm(req: NextApiRequest, res: NextApiResponse) {
   }
 
   try {
+    const passesSubmissionFilter = submissionFilter
+      ? await submissionFilter(req)
+      : true
+    if (!passesSubmissionFilter) {
+      // trick spammers into thinking their form submitted successfully
```
mwickett, line 108: 🪤

packages/marketo-form/form/index.tsx
```diff
@@ -170,6 +174,23 @@ const Form = ({
     },
   })
 
+  // Reset form if initialValues changes since defaultValues is cached on the
+  // first render. This handles the situation where initialValues is updated
+  // after the form is rendered, primarily when using router state which is
+  // resolved _after_ the initial render for SSG pages.
```
BRKalow, line 180: helpful comment! 💡

packages/marketo-form/server/api-routes.ts
```diff
@@ -39,9 +58,35 @@ async function getForm(req: NextApiRequest, res: NextApiResponse) {
...
+  // running e2e tests locally
   if (process.env.E2E_TESTS === 'true') {
+    return true
+  }
+
+  // running e2e tests against a Vercel preview
+  if (
+    'input' in req.body &&
+    req.body.input.length > 0 &&
+    req.body.input[0].leadFormFields.email === 'daniela.rod@example.com'
```
BRKalow, line 71: I'm guessing this is an arbitrary fixture email that's used? 😄

BRKalow (approved):
👏

packages/marketo-form/server/api-routes.ts
dstaley: It is! I'm not in love with this solution, but it's the best I could come up with given the constraints that we want preview deployments to be able to submit to Marketo, and we want E2E tests to run against preview deployments.

I could probably check the UA for `puppeteer` or something, but I'll explore that later.

alexcarpenter (approved):
Awesome work! 👏🏼

PR #634, opened 2022-06-30
feat(motion-config): use custom features configuration

dstaley:
🎟️ [Asana Task]()

---

## Description

This PR replaces `MotionConfig`'s usage of `domMax` with a custom feature set (basically `domMax - drag`) to reduce bundle size. This shouldn't break the intended usage of `MotionConfig`, but it also supports manual dynamic loading via the following:

```ts
import('@hashicorp/react-motion-config/features').then((mod) => mod.default)
```

Depends on: https://github.com/hashicorp/web-platform-packages/pull/56

### PR Checklist 🚀

### DIFF
--- a/.changeset/sour-snails-remain.md
+++ b/.changeset/sour-snails-remain.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-motion-config': minor
+---
+
+Use custom features configuration to reduce bundle size
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,7 +34,7 @@
         "@hashicorp/platform-cms": "^0.3.0",
         "@hashicorp/platform-code-highlighting": "0.1.2",
         "@hashicorp/platform-docs-mdx": "0.1.3",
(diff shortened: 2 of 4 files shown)


alexcarpenter (commented):
Guessing the features need to be loaded within the app level to be lazy loaded properly? 

Can we modify MotionConfig to expose features:

```
<MotionConfig features={import('@hashicorp/react-motion-config/features').then((mod) => mod.default)}>
</MotionConfig>
```

Just trying to reduce any boilerplate needed for implementation and keep updates needed in the future scoped to the motion-config package.

dstaley:
So actually, since Dev Portal [dynamically loads Notifications](https://github.com/hashicorp/dev-portal/pull/636/files#diff-88a36bc7a53bfa58c6f451464798d631d785160dfacfdaf9479be7a0b1d7e5e5R29-R35), the dynamic import in `MotionConfig` seems to work as intended, so I don't think we need the import directly in `_app`.

PR #656, opened 2022-07-19
feat: introduce notification types

alexcarpenter:
## Description

Based on whether or not the notification contains interactive content, render the notification with `role="alert"` for toasts and `role="alerDialog"` for dialogs.

This is based off the work HDS team has done with their [toast implementations](https://design-system-components-hashicorp.vercel.app/components/toast).

And confirmation from HDS team https://hashicorp.slack.com/archives/C7KTUHNUS/p1657818549189699?thread_ts=1657652997.846709&cid=C7KTUHNUS

### Usage

```tsx
notification() // defaults to toast type
notification.toast() // type = toast with role = alert
notification.dialog() // type = dialog with role = alertDialog
```

https://user-images.githubusercontent.com/825855/179763363-611d8b2c-0f60-4864-b44e-f0dff14a8951.mp4

### PR Checklist 🚀

### DIFF
--- a/packages/notification/components/notifications/index.tsx
+++ b/packages/notification/components/notifications/index.tsx
@@ -28,6 +28,12 @@ export const Notifications = ({
     >
       <AnimatePresence initial={false}>
         {notifications.map((n) => {
+          /**
+           * Define the role attribute based on toast type.
+           * toast.dialog() should be used in the case that
+           * the notification contians interactive content.
+           */
+          const role = n.type === 'toast' ? 'alert' : 'alertDialog'
           return (
             <m.div
               key={n.id}
@@ -39,7 +45,7 @@ export const Notifications = ({
                 scale: reducedMotion ? 1 : 0.75,
                 transition: { duration: 0.2 },
               }}
-              role="status"
+              role={role}
               aria-live="polite"
             >
               {renderNotification(n.message, n)}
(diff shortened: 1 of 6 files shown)


alexcarpenter:
In use: https://github.com/hashicorp/hashicorp-www-next/pull/2923

packages/notification/components/notifications/index.tsx
```diff
@@ -39,7 +45,7 @@ export const Notifications = ({
                 scale: reducedMotion ? 1 : 0.75,
                 transition: { duration: 0.2 },
               }}
-              role="status"
+              role={role}
```
ashleemboyer, line 48: (Question 🙋🏻) Is it possible to auto-detect the role based on whether or not the notification contains interactive content?

alexcarpenter: The thinking here was following the same reason we have explicit [Button](https://github.com/hashicorp/dev-portal/tree/main/src/components/button) and [ButtonLink](https://github.com/hashicorp/dev-portal/tree/main/src/components/button-link) components vs a single Button component that changes based on the attributes provided.

packages/notification/types.ts
```diff
@@ -4,6 +4,7 @@ export type Renderable = JSX.Element | string | null
 
 export type ValueFunction<TValue, TArg> = (arg: TArg) => TValue
 export type ValueOrFunction<TValue, TArg> = TValue | ValueFunction<TValue, TArg>
+export type NotificationAsType = 'toast' | 'dialog'
```
dstaley, line 7: nit: I think this type name would be better as `NotificationType`.

dstaley (approved):

packages/notification/types.ts
alexcarpenter: Updated https://github.com/hashicorp/react-components/pull/656/commits/a7e73d52f83d5a43f7989df4d832110033e15e75

PR #663, opened 2022-07-25
fix: prevent ol styles applying to nested list items

alexcarpenter:

---

## Description

Issue raised in slack that ol styles were applying to nested ul list items.

https://hashicorp.slack.com/archives/C5FSPUGDS/p1658506577270079

Fix implemented was to prevent applying li styles to all nested children within ol and only apply to their top level li elements.

**Before:**

CleanShot 2022-07-25 at 09 47 06@2x

**After:**

CleanShot 2022-07-25 at 09 42 28@2x

### PR Checklist 🚀

### DIFF
--- a/.changeset/light-teachers-behave.md
+++ b/.changeset/light-teachers-behave.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-content': patch
+---
+
+fix ol styles leaking to nested ul list items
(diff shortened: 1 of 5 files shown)


BRKalow (approved):
👍 makes sense to me! Thanks for the fix. Want to add a changeset here as well?

alexcarpenter:
Tested on `/docs/consul/usage/clusters` which is currently rendering incorrectly with the fix

CleanShot 2022-07-25 at 14 08 13@2x

PR #665, opened 2022-07-27
chore(docs-sidenav): improve key uniqueness

thiskevinwang:
🎟️ [Asana Task]()

---

## Description

This is only a perceived problem on some Terraform.io previews that are temporarily merging Cloud-Docs and Enterprise nav-data. Given that some nodes have the same titles and such, the addition of keys here increases the uniqueness and ensures that there are no duplicate key

### PR Checklist 🚀

### DIFF
--- a/packages/docs-sidenav/index.tsx
+++ b/packages/docs-sidenav/index.tsx
@@ -195,7 +195,7 @@ function NavTree({ baseRoute, content }: NavTreeProps) {
         if ('title' in item && 'href' in item) {
           return (
             <DirectLink
-              key={item.title + item.href}
+              key={item.title + item.href + idx}
               title={item.title}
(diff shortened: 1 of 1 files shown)


packages/docs-sidenav/index.tsx
```diff
@@ -195,7 +195,7 @@ function NavTree({ baseRoute, content }: NavTreeProps) {
         if ('title' in item && 'href' in item) {
           return (
             <DirectLink
-              key={item.title + item.href}
+              key={item.title + item.href + idx}
```
thiskevinwang, line 198: > **Note**: we are already relying on the index for `heading` and `divider` node types

BRKalow (approved):

PR #668, opened 2022-07-28
feat(docs-page): local version faking

thiskevinwang:
# Description

This enables local version-faking when using the `FS` content loader.

You can force a "version" into state by inserting a version like `v1.1.x` or `v202207-1` after the subpath in the URL.

### DIFF
--- a/packages/docs-page/server/loaders/file-system.ts
+++ b/packages/docs-page/server/loaders/file-system.ts
@@ -6,6 +6,8 @@ import { GetStaticPropsContext } from 'next'
 import { getPathsFromNavData } from '../get-paths-from-nav-data'
 import { resolveNavData } from '../resolve-nav-data'
 
+import { stripVersionFromPathParams } from '../../util'
+
 import { DEFAULT_PARAM_ID } from '../consts'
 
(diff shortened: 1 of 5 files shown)


packages/docs-page/server/loaders/file-system.ts
```diff
@@ -18,7 +20,9 @@ interface FileSystemLoaderOpts extends DataLoaderOpts {
   navDataFile: string
   localContentDir: string
   mainBranch?: string // = 'main',
-  remarkPlugins?: ((params?: ParsedUrlQuery) => Pluggable[]) | Pluggable[]
+  remarkPlugins?:
+    | ((params?: ParsedUrlQuery, version?: string) => Pluggable[])
```
thiskevinwang, line 24: @BRKalow What do you think about making this callback func send a `ctx` object, with fields like `ctx.params` & `ctx.version`...

(Would be a Major/breaking change)

BRKalow: Hmm, my gut says no for now as we're just creating more work for ourselves to update any existing usages. 💭

BRKalow (approved):

PR #674, opened 2022-08-02
feat: refactor version-alert text

thiskevinwang:
This updates the version-alert to use the `display` value that is returned from the ContentAPI

CleanShot 2022-08-02 at 16 21 16@2x

CleanShot 2022-08-02 at 16 21 45@2x

### DIFF
--- a/packages/docs-page/components/version-alert/index.tsx
+++ b/packages/docs-page/components/version-alert/index.tsx
@@ -5,25 +5,24 @@ import {
   removeVersionFromPath,
 } from '../version-select/util'
 import s from './style.module.css'
-import useIsMobile from '../../use-is-mobile'
 
-export default function VersionAlert({ product }) {
+export interface VersionAlertProps {
+  tag: string
+  text: string
+}
+export default function VersionAlert({ tag, text }: VersionAlertProps) {
   const router = useRouter()
-  const isMobile = useIsMobile()
   const versionInPath = getVersionFromPath(router.asPath)
 
   if (!versionInPath) return null
 
   return (
     <div className={s.wrapper}>
+      {/* @ts-expect-error: explicitly not passing `product` */}
       <Alert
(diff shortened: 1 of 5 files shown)


packages/docs-page/components/version-alert/index.tsx
```diff
@@ -5,25 +5,24 @@ import {
   removeVersionFromPath,
 } from '../version-select/util'
 import s from './style.module.css'
```
thiskevinwang, line 7: This cuts down some logic in this component, and makes it closer to a purely presentational component

packages/docs-page/components/version-alert/version-alert.test.tsx
```diff
@@ -20,38 +20,27 @@ describe('<VersionAlert />', () => {
...
+      } as unknown as Router
+    })
+
+    const { getByTestId } = render(
+      <VersionAlert tag="old version" text="Some Text" />
+    )
+
+    expect(getByTestId('tag')).toHaveTextContent('old version')
+    expect(getByTestId('text')).toHaveTextContent('Some Text')
+  })
```
thiskevinwang, line 45: Cleaned up tests due to slimmed down logic

thiskevinwang: There are only one thing to test for now — if `versionInPath` is present, or not.

packages/docs-page/index.tsx
```diff
@@ -69,6 +69,11 @@ export const DocsPageInner: FunctionComponent<DocsPageInnerProps> = ({
     versionInPath &&
     versions?.find((v) => v.isLatest)?.version === versionInPath
 
+  //
```
BRKalow, line 72:
```suggestion
```

BRKalow (approved):

PR #694, opened 2022-08-15
feat(component): `<Card />`

EnMod:
🎟️ [Asana Task](https://app.asana.com/0/1202026147841446/1202738878282259/f)
🔍 [Preview Link](https://react-components-git-nq-wpl-card-component-hashicorp.vercel.app/components/card)

---

## Description

This PR implements a new `<Card />` component, alongside sub-components consumers can use to compose its contents.

### PR Checklist 🚀

### DIFF
--- a/packages/card/docs.mdx
+++ b/packages/card/docs.mdx
@@ -154,6 +154,9 @@ Cards are used to promote other marketing pages.
           'Display 5 - 4 lines with character limit of 130. Leo mauris fermentum pharetra blandit tellus euismod…'
         }
       </Card.Heading>
+        <p style={fontSize: '16px'}>{
+          'Display 5 - 4 lines with character limit of 130. Leo mauris fermentum pharetra blandit tellus euismod…'
+        }</p>
     </Card.Content>
   </Card>
 </div>
--- a/packages/card/docs.mdx
+++ b/packages/card/docs.mdx
@@ -135,6 +135,11 @@ Cards are used to promote other marketing pages.
   <Card link="https://hashicorp.com">
     <Card.Content>
       <Card.Meta items={['August 15, 2022', 'Category']} />
+      <Card.Heading>
+        {
+          'Display 5 - 4 lines with character limit of 130. Leo mauris fermentum pharetra blandit tellus euismod…'
+        }
+      </Card.Heading>
       <Card.Description>
         {
           'Body small - 4 lines with character limit of 170. Leo mauris fermentum pharetra blandit tellus euismod pharetra. Duis egestas volutpat dolor…'
--- a/packages/card/docs.mdx
+++ b/packages/card/docs.mdx
@@ -7,7 +7,7 @@ Cards are [placeholder description here].
 <LiveComponent>
   {`
 <>
-  <div style={{display: 'flex', gap: '32px'}}>
+  <div style={{display: 'grid', gridTemplateColumns: '1fr 1fr 1fr', gap: '32px'}}>
     <Card
       link="https://hashicorp.com"
       thumbnail={
@@ -48,7 +48,14 @@ Cards are [placeholder description here].
   Light theme
 </h3>
 
-<div style={{ display: 'flex', gap: '32px', marginBottom: '32px' }}>
+<div
+  style={{
+    display: 'grid',
+    gridTemplateColumns: '1fr 1fr 1fr',
+    gap: '32px',
+    marginBottom: '32px',
+  }}
+>
   <Card
     link="https://hashicorp.com"
     heading="Display 5 - 4 lines with character limit of 130. Leo mauris fermentum pharetra blandit tellus euismod…"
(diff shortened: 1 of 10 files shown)


packages/card/style.module.css
```diff
@@ -73,5 +73,5 @@
   align-items: center;
   justify-content: flex-end;
   color: var(--wpl-neutral-200);
-  padding: 20px;
+  padding: 24px;
```
alexcarpenter, line 76: We may want padding-top: 12px here from what I am seeing in the design.

EnMod: Hmm...perhaps even `8px`? Noticing that is the gap between the content and the outer border of the arrow icon in the design.

Pushed a commit so you can compare. I also removed bottom padding on `<Content />` to reduce the spacing further to match.

alexcarpenter: well can content include 24px padding on each side and cta has now padding top?

EnMod: Hmm....`<Content />` can't have `padding-bottom`, from what I can see, if CTA gets `8px` of `padding-top`.  Otherwise, the space between `<Content />` and the CTA grows too large vs the design 🤔 

Unless `<Content />` gets `4px` of `padding-bottom`, then the CTA gets `4px` of `padding-top`? But in that case the spacing between CTA and other possible children then becomes too small?🤔

packages/card/index.test.js
```diff
@@ -0,0 +1,28 @@
+import { render } from '@testing-library/react'
+import Card from '.'
+
+describe('<Card />', () => {
+  it('should render the provided React children', () => {
```
alexcarpenter, line 5: Some good basic test would be testing the card with props version renders each piece of content, so heading is in the document, description is in the document, the link is correct, etc.

packages/card/docs.mdx
```diff
@@ -0,0 +1,205 @@
+---
+componentName: 'Card'
+---
+
+Cards are used to promote other marketing pages.
```
mwickett, line 5: turbo nit: this feels like one possible use for cards, but this description could be a bit more nuanced.

mwickett (commented):
Looking good! Left a super nitpick comment about the description of the component in `docs.mdx`.

Really stoked to see us roll this out.

packages/card/docs.mdx
EnMod: Agreed! This description is mainly just a rewording of the one by design for now.  It's worth workshopping a more holistic description here :+1:

packages/card/index.test.js
EnMod: Yep, good call.  Additionally, perhaps some tests checking for the presence of composable children's content if added?  Tests for that can check to see if any props were passed to the `<Card />`; given how we've set it up, that test should fail if there are props being passed _and_ composable children being used.  

The TypeScript typing has been set up to cover this though, so this might be a case of testing the implementation details (which we probably want to avoid?)

packages/card/docs.mdx
EnMod: @mwickett Updated the description here:
https://github.com/hashicorp/react-components/pull/694/commits/5d5a9cb556ef5b2334c62622cfaaf1f53355469d

It's more detailed but I still think it's missing something 🤔 Thoughts?

packages/card/index.test.js
EnMod: @alexcarpenter Wrote a suite of prop render tests here:
https://github.com/hashicorp/react-components/pull/694/commits/14a448c879f47fbc19aa21486cca5bf5ba7fea53

In writing these tests I was primarily pulling from conventions found elsewhere in our codebase; any feedback on methods used here is appreciated 🙏

alexcarpenter: Things are looking good! Recorded a bit of a brain dump: https://hashicorp.slack.com/archives/C031LA4KS4S/p1660856664941599

Might be helpful to have a test ensuring the link is in the card too.

packages/card/index.test.js
```diff
@@ -69,55 +52,33 @@ describe('<Card />', () => {
...
-        heading="Example heading"
-      />
-    )
-
-    const thumbnailElement = screen.getByTestId('wpl-card-thumbnail')
-    expect(thumbnailElement).toBeInTheDocument()
-    expect(getByAltText(thumbnail.alt).tagName).toBe('IMG')
+    const { getByAltText } = render(<Card {...defaultProps} />)
+
+    expect(getByAltText(defaultProps.thumbnail.alt).tagName).toBe('IMG')
```
alexcarpenter, line 57: Nice update with getByAltText selector here 👏🏼

packages/card/index.test.js
```diff
@@ -0,0 +1,151 @@
...
+    const arbitraryText = 'Arbitrary text'
+
+    render(
+      <Card link={defaultProps.link} heading={defaultProps.heading}>
+        {arbitraryText}
+      </Card>
+    )
+
+    screen.getByText(arbitraryText)
+    expect(screen.queryByText(defaultProps.heading)).not.toBeInTheDocument()
```
EnMod, line 133: I tried using `getByText()` here and that always failed because the heading component is not supposed to be in the document, so it never was able to be found.

That's when I found `queryByText()`! The difference? `getByText()` will fail the test if it can't find the text, but `queryByText()` will return `null` instead. 

So, if `queryByText()` finds a heading element here, the test will fail as intended.

packages/card/index.test.js
EnMod: Just pushed tests covering the link, and also setting heading levels in `<Card.Heading />`. I've replied in Slack on that brain dump, some great insights there!

pbortnick (approved):
LGTM ✨

pbortnick (approved):

PR #705, opened 2022-08-19
feat(subnav): adds `menuText` prop

alexcarpenter:
🔍 [Preview Link](https://react-components-git-acsubnav-menu-text-hashicorp.vercel.app/components/subnav)

---

## Description
This PR updates the `Subnav` component such that it now takes a custom menu text prop to enable Menu text translation in [i18n microsites](https://github.com/hashicorp/internationalized-microsites/pull/58).

### PR Checklist 🚀

### DIFF
--- a/packages/subnav/partials/MenuItemsOverflow/style.module.css
+++ b/packages/subnav/partials/MenuItemsOverflow/style.module.css
@@ -16,7 +16,7 @@
   position: absolute;
   right: 0;
   top: 100%;
-  min-width: 256px;
+  width: 256px;
   z-index: 1;
 
   &.isCollapsed {
--- a/packages/subnav/partials/MenuItemsOverflow/style.module.css
+++ b/packages/subnav/partials/MenuItemsOverflow/style.module.css
@@ -16,7 +16,7 @@
   position: absolute;
   right: 0;
   top: 100%;
-  width: 256px;
+  min-width: 256px;
   z-index: 1;
 
   &.isCollapsed {
--- a/.changeset/late-planes-hear.md
+++ b/.changeset/late-planes-hear.md
@@ -2,4 +2,4 @@
 '@hashicorp/react-subnav': minor
 ---
 
-Adds menuText prop and min-width to dropdown
+Adds menuText prop
--- a/.changeset/late-planes-hear.md
+++ b/.changeset/late-planes-hear.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-subnav': minor
+---
+
+Adds menuText prop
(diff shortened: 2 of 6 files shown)


zchsh (approved):
🙌 This looks great to me 👍

nandereck:
@zchsh - mind giving this another look? I added one more change to this PR that makes the `Subnav` component usable on i18n microsites.

packages/subnav/partials/MenuItemsOverflow/style.module.css
```diff
@@ -16,7 +16,7 @@
   position: absolute;
   right: 0;
   top: 100%;
-  width: 256px;
+  min-width: 256px;
```
zchsh, line 19: 💭 One potential challenge here might be on small viewports.

I think the `256px` value here was set as a strict thing rather than a `min` to try to more explicitly express that the dropdown can't get much wider than `256px` on small viewports without causing breakage:

CleanShot 2022-08-22 at 17 19 55@2x

That being said, maybe this is too prescriptive, and I suppose if an element within the dropdown is wider than `256px`, we might end up seeing breakage either way... and using `min-width` does seem to offer flexibility to do wider things on larger viewports with media queries... and I trust there's some i8n-related rationale here too...  so I'm on board with this, just wanted to flag in case that concern is relevant! ☺️

zchsh (approved):

packages/subnav/partials/MenuItemsOverflow/style.module.css
nandereck: Gah, good catch. Just to confirm I understand though, setting the `min-width` to 256px means that the dropdown will grow in width depending on the content. The breakage you're referring occurs when the dropdown expands to a width (due to longer strings within it) that exceeds that of the viewport. Does that sound right?

Ex:
Screen Shot 2022-08-22 at 5 40 21 PM

The context here is that I'm swapping out the custom nav used on the internationalized microsites with the `Subnav` component. Unfortunately, the text for the French CTA (see above) is pretty long so the CTA is overflowing out of the dropdown and off of the page. Given the issue you've flagged and the fact that the French text _seems_ like it should be shorter, I'll try to track down someone on the i18n side to see if it makes sense to come up with shorter copy.

In the meantime, I'll revert my commit and merge Alex's approved changes. Thanks for taking a look!

zchsh: Oooh yeah dang these are some interesting edge cases... for that button, I _think_ the ideal scenario would have the button text wrap? 🤔 That's a hot take though, I'm not sure!

PR #719, opened 2022-09-07
Use `useErrorPageAnalytics` hook on 404 page

nandereck:
🎟️ [Asana Task](https://app.asana.com/0/1202791227659279/1202865459293217/f)

---

## Description
It was recently discovered that, across our various web properties, we make the following call:

`window.analytics.track(window.location.href, { ... } )`

The first parameter of the `track` method is used downstream in our various data warehouses to create a new table. Since we're using `window.location.href`, we're creating a table for _every_ URL that 404s. This is not ideal for analytics.

## Solution
This PR leverages the new `useErrorPageAnalytics` hook (see [this PR](https://github.com/hashicorp/web-platform-packages/pull/86)) for pushing 404-related data to Segment. This PR replaces the `window.analytics.track` call on the 404 page with said hook.

### PR Checklist 🚀

### DIFF
--- a/.changeset/stupid-lions-grab.md
+++ b/.changeset/stupid-lions-grab.md
@@ -2,4 +2,4 @@
 '@hashicorp/react-error-view': minor
 ---
 
-Uses the useErrorPageAnalytics hook
+Replaces track event name with static value.
--- a/.changeset/stupid-lions-grab.md
+++ b/.changeset/stupid-lions-grab.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-error-view': minor
+---
+
+Uses the useErrorPageAnalytics hook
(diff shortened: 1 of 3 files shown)


.changeset/stupid-lions-grab.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hashicorp/react-error-view': minor
+---
+
+Uses the useErrorPageAnalytics hook
```
alexcarpenter, line 5: Doesn't look like we are leveraging the `useErrorPageAnalytics` hook on this implementation.

```suggestion
Replaces track event name with static value.
```

alexcarpenter (commented):
@BRKalow are we looking to collect a different track event here with Click-side Redirect vs 404 here?

BRKalow:
It seems like we are deliberately collecting a different event here, yeah. 🤔 It would in theory provide insight if we have internal links pointing at redirects, which we would want to fix in content.

alexcarpenter (approved):

.changeset/stupid-lions-grab.md
nandereck: Oops!

PR #731, opened 2022-09-28
Create `RelatedContent` WPL component

nandereck:
🎟️ [Asana Task]()

---

## Description

### PR Checklist 🚀


packages/related-content/types.ts
```diff
@@ -0,0 +1,13 @@
...
+export interface RelatedContentProps {
+  appearance?: 'light' | 'dark'
+  headline: string
+  description?: string
+  cards: Array<CardProps>
+  cta?: {
+    title: string
+    url: string
+    variant: ThemeVariant
+  }
```
alexcarpenter, line 12: Can we extend StandaloneLink props here vs custom with Button ThemeVariant?

packages/related-content/style.module.css
```diff
@@ -0,0 +1,63 @@
+.wrapper {
+  composes: g-grid from global;
+  composes: g-grid-container from global;
+  &.dark {
+    background-color: var(--wpl-neutral-900);
+  }
+}
```
alexcarpenter, line 7: Is it up to the consumer of the component to include the vertical padding of this component?

If so, I would leave the background color up to their decision as well. Since they'd have to apply the same background color you are applying here as well in dark appearance.

packages/related-content/props.js
```diff
@@ -0,0 +1,85 @@
...
+    type: 'string',
+    required: false,
+  },
+  link: {
+    type: 'string',
+    description: 'A link followed when the card is clicked.',
+    required: true,
+    testValue: 'https://hashicorp.com',
+  },
+}
```
alexcarpenter, line 80: 💭 Curious about including the Card props here. Thoughts on linking to the Card page documentation vs maintaining two spots of card docs?

packages/related-content/props.js
```diff
@@ -0,0 +1,85 @@
...
+    required: false,
+    testValue: ['August 15, 2022', 'Category'],
+  },
+  appearance: {
+    description: 'Styles the card with either a light or dark theme.',
+    type: 'string',
+    required: false,
+    options: ['light', 'dark'],
+    testValue: 'light',
+  },
```
alexcarpenter, line 63: If we decide to keep the card docs here, we do not use the appearance for the card object, it takes the appearance from RelatedContent component prop.

packages/related-content/types.ts
```diff
@@ -0,0 +1,13 @@
+import type { CardProps } from '@hashicorp/react-card/types'
+import type { ThemeVariant } from '@hashicorp/react-button/types'
+export interface RelatedContentProps {
+  appearance?: 'light' | 'dark'
+  headline: string
+  description?: string
+  cards: Array<CardProps>
```
alexcarpenter, line 7: 💭 Possibly adjust CardProps to exclude appearance usage here for clarity.

packages/related-content/props.js
nandereck: That's way better of an idea. 👍

packages/related-content/props.js
nandereck: 🔪 [here](https://github.com/hashicorp/react-components/pull/731/commits/8f7ee6a9f5540c9e4788bc7a1e38cfb3fcc89ae6)

packages/related-content/style.module.css
nandereck: Added vertical padding. 👍

packages/related-content/types.ts
nandereck: Updated [here](https://github.com/hashicorp/react-components/pull/731/commits/d43f0bff7a331468ebce57d688fd5b13a2dfd4e0).

packages/related-content/style.module.css
```diff
@@ -0,0 +1,70 @@
...
+  display: grid;
+  grid-gap: 32px;
+  grid-template-columns: repeat(var(--columns), 1fr);
+  margin-bottom: 32px;
+
+  @media (--medium) {
+    --columns: 2;
+  }
+  @media (--large) {
+    --columns: 4;
```
alexcarpenter, line 64: Just to confirm expectations, if there are three cards provided, we still render 4 columns across vs 3 columns across?

alexcarpenter: CleanShot 2022-10-03 at 10 42 25@2x

Bottom example rendering 3 columns vs 4.

nandereck: > Just to confirm expectations, if there are three cards provided, we still render 4 columns across vs 3 columns across?

My understanding is that this component can only render 4 cards.

alexcarpenter: Interesting, so to make use of RelatedContent, folks will need to provide at least 4 cards. Or it's ok that less than 4 cards are provided, they just will still render within 4 columns? I could see the need to use less based on some of content the content I have seen on .io sites.

If we've decided that is a requirement, what do you think about adding a length check to the cards array that at least 4 cards are provided then or update types to reflect 4 cards?

```tsx
cards: [RelatedContentCardProps, RelatedContentCardProps, RelatedContentCardProps, RelatedContentCardProps]
```

alexcarpenter: FWIW also seeing less than and more than 4 related resources being used in the work @EnMod is doing with SI Partners.

- https://hashicorp-www-next-git-nqsi-update-template-hashicorp.vercel.app/partners/systems-integrators/adfinis
- https://hashicorp-www-next-git-nqsi-update-template-hashicorp.vercel.app/partners/systems-integrators/amazic

nandereck: >> Interesting, so to make use of RelatedContent, folks will need to provide at least 4 cards. Or it's ok that less than 4 cards are provided, they just will still render within 4 columns? I could see the need to use less based on some of content the content I have seen on .io sites.

FWIW - I think this component should allow for multiple cards and not be limited to only 4. I wonder if it makes sense to enable that and make adjustments later. Especially if we're already seeing deviations from the initial requirements.

alexcarpenter (approved):
👍🏼

packages/related-content/style.module.css
EnMod: @nandereck I agree: I think it makes sense to go with how it's currently implemented (render with 4 columns regardless of count, do not limit to 4 cards minimum). Especially with deviations already existing. 

As an aside, this implementation is aligned with how local related resources components are currently implemented, an example being on [our thank you page](https://www.hashicorp.com/thank-you) :+1:

EnMod (approved):
💯

PR #748, opened 2022-10-05
feat: Add `<ProductBadge />` support to `<Card />` component

nandereck:
🎟️ [Asana Task](https://app.asana.com/0/1202802428287999/1203049731246701/f)
🔍 [Preview Link](https://react-components-git-nels-cardadd-product-badg-3c55a7-hashicorp.vercel.app/components/card)

---

## Description

This PR enables `<ProductBadge />` support to the `<Card />` component. Couple of call-outs:
- The spec for the `Card` component displays the `hasDot` variant of `ProductBadge`. Given that and a preference to be a bit restrictive, we force that variant in the code.
- Similarly, the code does not allow `ProductBadge` to be a composable child.

## Validation Steps
- Navigate to the [preview link](https://react-components-git-nels-cardadd-product-badg-3c55a7-hashicorp.vercel.app/components/card)
- Add the following prop to any of the cards in the Code Editor:
`productBadges={{badges: [{ productName: 'waypoint'}]}}`
- Verify that the Product Badges render properly
- Review the Use Case Cards in the Examples section

### PR Checklist 🚀

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
     "@hashicorp/flight-icons": "^2.5.0",
     "@hashicorp/mktg-global-styles": "^4.2.0",
     "@hashicorp/mktg-logos": "^1.3.2",
+    "@hashicorp/react-product-badge": "^0.1.1",
     "@hashicorp/sentinel-embedded": "^0.0.14",
     "classnames": "^2.3.1",
     "formik": "^2.2.9",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,6 @@
     "@hashicorp/flight-icons": "^2.5.0",
     "@hashicorp/mktg-global-styles": "^4.2.0",
     "@hashicorp/mktg-logos": "^1.3.2",
-    "@hashicorp/react-product-badge": "^0.1.1",
     "@hashicorp/sentinel-embedded": "^0.0.14",
     "classnames": "^2.3.1",
     "formik": "^2.2.9",
--- a/packages/card/index.tsx
+++ b/packages/card/index.tsx
@@ -37,6 +37,7 @@ function Card(props: CardProps) {
           <Content>
             {meta && meta.length > 0 ? <Meta items={meta} /> : null}
             <Heading>{heading}</Heading>
+            {description ? <Description>{description}</Description> : null}
             {productBadges &&
             productBadges?.badges &&
             productBadges?.badges?.length > 0 ? (
@@ -45,7 +46,6 @@ function Card(props: CardProps) {
                 appearance={appearance}
               />
             ) : null}
-            {description ? <Description>{description}</Description> : null}
           </Content>
         </>
       )}
--- a/packages/card/index.tsx
+++ b/packages/card/index.tsx
@@ -13,7 +13,6 @@ import type {
   DescriptionProps,
 } from './types'
 import s from './style.module.css'
-import React from 'react'
 
 function Card(props: CardProps) {
   const {
@@ -118,14 +117,13 @@ function ProductBadges({ badges, appearance = 'light' }: ProductBadgesProps) {
     <div className={s.productBadges}>
       {badges.map((badge, stableIdx) => {
         return (
-          // eslint-disable-next-line react/no-array-index-key
-          <React.Fragment key={stableIdx}>
-            <ProductBadge
-              appearance={appearance}
-              productName={badge.productName}
-              hasDot={true}
-            />
-          </React.Fragment>
+          <ProductBadge
+            // eslint-disable-next-line react/no-array-index-key
+            key={stableIdx}
+            appearance={appearance}
+            productName={badge.productName}
+            hasDot={true}
+          />
         )
       })}
     </div>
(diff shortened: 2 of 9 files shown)


packages/card/index.tsx
```diff
@@ -33,6 +37,14 @@ function Card(props: CardProps) {
...
             {meta && meta.length > 0 ? <Meta items={meta} /> : null}
             <Heading>{heading}</Heading>
+            {productBadges &&
+            productBadges?.badges &&
+            productBadges?.badges?.length > 0 ? (
+              <ProductBadges
+                badges={productBadges.badges}
+                appearance={appearance}
+              />
+            ) : null}
```
alexcarpenter, line 47: Looks like badges should display below the description.

CleanShot 2022-10-05 at 17 01 24@2x

packages/card/index.tsx
```diff
@@ -101,6 +113,25 @@ function Heading({ as: Component = 'h2', children }: HeadingProps) {
   )
 }
 
+function ProductBadges({ badges, appearance = 'light' }: ProductBadgesProps) {
```
alexcarpenter, line 116: Cards can only render productBadges with the dot and secondary theme?

package.json
```diff
@@ -5,6 +5,7 @@
     "@hashicorp/flight-icons": "^2.5.0",
     "@hashicorp/mktg-global-styles": "^4.2.0",
     "@hashicorp/mktg-logos": "^1.3.2",
+    "@hashicorp/react-product-badge": "^0.1.1",
```
dstaley, line 8: I don't think we want this dependency installed at the root. You can fix this by running the following in the root directory:

```
npm uninstall @hashicorp/react-product-badge
npm install @hashicorp/react-product-badge --workspace=@hashicorp/react-card
```

That will ensure that consumers who install `@hashicorp/react-card` get the `@hashicorp/react-product-badge` dependency.

packages/card/index.tsx
nandereck: Oops - you're right. [Moved](https://github.com/hashicorp/react-components/pull/748/commits/fbb86263b5577426c921f2bc54d978c61a3afd48)!

packages/card/index.tsx
```diff
@@ -101,6 +113,25 @@ function Heading({ as: Component = 'h2', children }: HeadingProps) {
...
+  return (
+    <div className={s.productBadges}>
+      {badges.map((badge, stableIdx) => {
+        return (
+          // eslint-disable-next-line react/no-array-index-key
+          <React.Fragment key={stableIdx}>
```
pbortnick, line 122: ❓ Do we need `<React.Fragment>` here or can we add the key directly to `<ProductBadge>` ?

dstaley (changes requested):

packages/card/index.tsx
nandereck: See my comment above in the PR description.

Tangentially related, but `hasDot` is the `secondary` theme plus a dot versus having totally separate styles if that makes sense.

packages/card/index.tsx
pbortnick: If not, can we also delete the React import?

packages/card/index.tsx
alexcarpenter: Got it, sorry I missed that in the description!

Thoughts on simplifying the prop if we're being restrictive here.

```tsx
<Card
  productBadges={['vault', 'consul']}
/>
```

packages/card/index.tsx
nandereck: Oops - good catch. This is cruft. [Removed](https://github.com/hashicorp/react-components/pull/748/commits/3be3b6a09666b959e6e73906a7bb10533fd0bca1)!

package.json
nandereck: Gah, my bad! Will do!

nandereck: Commit [here](https://github.com/hashicorp/react-components/pull/748/commits/0d5334ef7c6a10c3f1cc292b992dd1e8f64b2b56).

dstaley (approved):

packages/card/index.tsx
```diff
@@ -34,6 +37,14 @@ function Card(props: CardProps) {
             {meta && meta.length > 0 ? <Meta items={meta} /> : null}
             <Heading>{heading}</Heading>
             {description ? <Description>{description}</Description> : null}
+            {productBadges &&
+            productBadges?.badges &&
+            productBadges?.badges?.length > 0 ? (
+              <ProductBadges
+                badges={productBadges.badges}
+                appearance={appearance}
```
alexcarpenter, line 45: Maybe this is expected, but noticed the badge color is the same as the cards hover background color when appearance is dark.

CleanShot 2022-10-05 at 17 23 11@2x

nandereck: Let me double check, but I'm not sure this is accounted for in either the `Card` or `ProductBadge` spec. 😕

packages/card/types.ts
```diff
@@ -38,6 +41,11 @@ export interface HeadingProps {
   children: string
 }
 
+export interface ProductBadgesProps {
+  badges: Array<ProductBadgeProps['productName']>
```
nandereck, line 45: @alexcarpenter - thoughts on this? I've actually never utilized lookup types. 🙃

alexcarpenter (approved):
LGTM!

PR #758, opened 2022-10-12
Enable `StandaloneLink` in `Actions`

nandereck:
🎟️ [Asana Task](https://app.asana.com/0/1202802428287999/1203018584186578/f)
🔍 [Preview Link](https://react-components-git-nels-introadd-standalone-fd8f07-hashicorp.vercel.app/components/actions)
🎨 [Figma](https://www.figma.com/file/7cYgDM618stjYUHDqAfRec/Components?node-id=572%3A1058)

---

## Description

This PR adds support for the `StandaloneLink` in `Actions`. The requirements are as follows:
- `Actions` supports all of the `Button` / `StandaloneLink` combinations in the [spec](https://www.figma.com/file/7cYgDM618stjYUHDqAfRec/branch/5GbGjYszP8XpIebLm64Sa9/Components?node-id=572%3A1058).
- The first CTA and second CTA are always the `primary` variation and `secondary` variation respectively.
- When both a `Button` and `StandaloneLink` render, the `StandaloneLink` always comes second.
Addition

The changeset includes the following:
- Adds `StandaloneLink` to the render method of `Actions`
- Changes the `url` prop for `Actions` to `href`. This was done to match the `<a>` attribute name.
- Removes the now cruft `variant` prop from CTAs in `Actions`
- Updates the docs and tests to account for the new variations
- Updates failing tests for `Intro` and `NextSteps`. This occurs due to the change of the `url` prop to `href`

## Validation Steps
- Navigate to [`Actions` in the preview](https://react-components-git-nels-introadd-standalone-fd8f07-hashicorp.vercel.app/components/actions)
- Under examples, verify that all of the variations match the Figma (see above)

### PR Checklist 🚀

### DIFF
--- a/.changeset/tidy-emus-hammer.md
+++ b/.changeset/tidy-emus-hammer.md
@@ -1,7 +1,0 @@
----
-'@hashicorp/react-actions': major
-'@hashicorp/react-intro': major
-'@hashicorp/react-next-steps': major
----
-
-Enable StandaloneLink support for Actions and update tests for Intro and NextSteps
--- a/.changeset/tidy-emus-hammer.md
+++ b/.changeset/tidy-emus-hammer.md
@@ -1,0 +1,7 @@
+---
+'@hashicorp/react-actions': major
+'@hashicorp/react-intro': major
+'@hashicorp/react-next-steps': major
+---
+
+Enable StandaloneLink support for Actions and update tests for Intro and NextSteps
--- a/packages/actions/index.tsx
+++ b/packages/actions/index.tsx
@@ -28,7 +28,7 @@ export default function Actions({
         if (isStandaloneLink) {
           return (
             <StandaloneLink
-              href={cta.url}
+              href={cta.href}
               theme={variant}
               appearance={appearance}
               onClick={cta?.onClick}
@@ -42,7 +42,7 @@ export default function Actions({
               // eslint-disable-next-line react/no-array-index-key
               key={index}
               title={cta.title}
-              url={cta.url}
+              url={cta.href}
               onClick={cta.onClick}
               size={size}
               theme={{
--- a/packages/actions/index.tsx
+++ b/packages/actions/index.tsx
@@ -21,8 +21,6 @@ export default function Actions({
       {ctas.map((cta, index) => {
         // The first CTA and second CTA should always
         // have the `primary` and `secondary` variations respectively.
-        // This const is confusingly named because `Button` and `StandaloneLink`
-        // use different prop names to trigger variations (`variant` and `theme`).
 
         const variant = index === 0 ? 'primary' : 'secondary'
         const isStandaloneLink = cta?.type === 'standalone-link'
--- a/packages/actions/index.tsx
+++ b/packages/actions/index.tsx
@@ -24,11 +24,21 @@ export default function Actions({
         // This const is confusingly named because `Button` and `StandaloneLink`
         // use different prop names to trigger variations (`variant` and `theme`).
 
-        const themeOrVariant = index === 0 ? 'primary' : 'secondary'
-        const isStandaloneLink = cta?.href && cta?.children
-        const isButton = cta?.title && cta?.url
+        const variant = index === 0 ? 'primary' : 'secondary'
+        const isStandaloneLink = cta?.type === 'standalone-link'
 
-        if (isButton) {
+        if (isStandaloneLink) {
+          return (
+            <StandaloneLink
+              href={cta.url}
+              theme={variant}
+              appearance={appearance}
+              onClick={cta?.onClick}
+            >
+              {cta.title}
+            </StandaloneLink>
+          )
+        } else {
           return (
             <Button
               // eslint-disable-next-line react/no-array-index-key
@@ -39,22 +49,10 @@ export default function Actions({
               size={size}
               theme={{
                 brand: theme,
-                variant: themeOrVariant,
+                variant: variant,
                 background: appearance === 'dark' ? 'dark' : undefined,
               }}
             />
-          )
-        }
-
-        if (isStandaloneLink) {
-          return (
-            <StandaloneLink
-              href={cta.href}
-              theme={themeOrVariant}
-              appearance={appearance}
-            >
-              {cta.children}
-            </StandaloneLink>
           )
         }
       })}
--- a/packages/actions/index.tsx
+++ b/packages/actions/index.tsx
@@ -34,7 +34,7 @@ export default function Actions({
               theme={variant}
               appearance={appearance}
               onClick={cta?.onClick}
-              data-testid="standaloneLink"
+              data-testid={`standaloneLink-${index}`}
             >
               {cta.title}
             </StandaloneLink>
@@ -53,7 +53,7 @@ export default function Actions({
                 variant: variant,
                 background: appearance === 'dark' ? 'dark' : undefined,
               }}
-              data-testid="button"
+              data-testid={`button-${index}`}
             />
           )
         }
--- a/packages/actions/index.tsx
+++ b/packages/actions/index.tsx
@@ -34,7 +34,6 @@ export default function Actions({
               // eslint-disable-next-line react/no-array-index-key
               key={index}
               title={cta.title}
-              linkType={undefined}
               url={cta.url}
               onClick={cta.onClick}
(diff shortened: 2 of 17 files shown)


nandereck:
@alexcarpenter - WIP PR here. Curious to get your thoughts on this. 

I _think_  this hits all the requirements (see the description above), is relatively understandable, and is backwards compatible with the existing `Actions` component.

There's also a high likelihood that I'm overlooking something important here too 😅 🙃

packages/actions/types.ts
```diff
@@ -38,7 +39,17 @@ export interface ActionsProps {
...
+   * requirement of rendering the Button first when data
+   * for both a Button and a StandaloneLink are provided.
+   * @see CtaProps and StandaloneLinkProps
+   */
+  ctas:
+    | [CtaProps]
+    | [StandaloneLinkProps]
+    | [CtaProps, CtaProps]
+    | [StandaloneLinkProps, StandaloneLinkProps]
+    | [CtaProps, StandaloneLinkProps]
```
nandereck, line 54: We enforce the order of the data here and thus the order of the CTAs.

packages/actions/index.tsx
```diff
@@ -18,23 +19,45 @@ export default function Actions({
...
-          />
-        )
+        // The first CTA and second CTA should always
+        // have the `primary` and `secondary` variations respectively.
+        // This const is confusingly named because `Button` and `StandaloneLink`
+        // use different prop names to trigger variations (`variant` and `theme`).
+
+        const themeOrVariant = index === 0 ? 'primary' : 'secondary'
+        const isStandaloneLink = cta?.href && cta?.children
+        const isButton = cta?.title && cta?.url
```
nandereck, line 29: This could very well be a code smell, but we're basically figuring out what component to render based on what props come in.

packages/actions/types.ts
alexcarpenter: I know the underlying components differ in props they accept, but I think when composing them we can make them the same.

Example:

```tsx
<Actions
  ctas={[
    {
      title: 'One',
      href: '/one'
    },
    {
      type: 'standalone-link',
      title: 'Two',
      href: '/two'
    }
  ]}
/>
```

CTA type:

```tsx
ctas: Array<{
  type: 'button' | 'standalone-link'
  title: string
  href: string
}>
```

and we default type to button.

This will remove the need to do the isButton and isStandalone link checks and just rely on the type passed in.

Thoughts?

packages/actions/index.tsx
```diff
@@ -18,23 +19,45 @@ export default function Actions({
...
+            />
+          )
+        }
+
+        if (isStandaloneLink) {
+          return (
+            <StandaloneLink
+              href={cta.href}
+              theme={themeOrVariant}
+              appearance={appearance}
```
alexcarpenter, line 55: Will also want to pass onClick handler here similar to the button.

```tsx
onClick={cta.onClick}
```

nandereck: Ah, yes. Done. 👍

nandereck: Ah, yes. Done 👍

packages/actions/index.tsx
```diff
@@ -18,23 +19,45 @@ export default function Actions({
...
-              background: appearance === 'dark' ? 'dark' : undefined,
-            }}
-          />
-        )
+        // The first CTA and second CTA should always
+        // have the `primary` and `secondary` variations respectively.
+        // This const is confusingly named because `Button` and `StandaloneLink`
+        // use different prop names to trigger variations (`variant` and `theme`).
+
+        const themeOrVariant = index === 0 ? 'primary' : 'secondary'
```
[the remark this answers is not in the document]
nandereck: Oh, right. I guess it doesn't really matter what I name the variable.

packages/actions/types.ts
nandereck: If I'm understanding correctly, yes I think this makes way more sense.

Is [this](https://github.com/hashicorp/react-components/pull/758/commits/f8d2272227df856ddd2eaf965f8ca4af40f718da) what you had in mind?

packages/actions/types.ts
```diff
@@ -19,6 +19,14 @@ interface CtaProps {
   onClick?: () => void
 }
 
+interface ButtonCtaProps extends CtaProps {
+  type?: 'button' | undefined
```
nandereck, line 23: Don't think I need `undefined` here if it's marked as optional.

packages/actions/types.ts
alexcarpenter: Yeah, I think we can update CtaProps to remove variant.

```ts
interface CtaProps {
  /**
   * The text that appears inside the button.
   */
  title: string
  /**
   * Where the button links to when clicked.
   */
  href: string
  /**
   * A function that will be called when the button is clicked.
   */
  onClick?: () => void
}

interface ButtonCtaProps extends CtaProps {
  type?: 'button'
}

interface StandaloneLinkCtaProps extends CtaProps {
  type: 'standalone-link'
}
```

alexcarpenter: I'd prefer we use href vs url since it matches the native link attribute. eg <a href="">...</a>.

In the future it'd be nice to move from url usage in button to href as well.

nandereck: > I'd prefer we use href vs url since it matches the native link attribute. eg <a href="">...</a>.

Totally onboard. Wouldn't changing `url` to `href` on `CtaProps` require either changes to models that are supplying data to `Actions` or data transformation after we query Dato? If so, do you think that is something that should be tackled in this PR? Or are you just flagging this more generally?

alexcarpenter: For clarity I am not saying updating button prop right now. But action ctas can use href vs url and we just pass the href to the url on button.

packages/actions/index.tsx
```diff
@@ -18,23 +19,42 @@ export default function Actions({
...
-              brand: theme,
-              variant: cta.variant || 'primary',
-              background: appearance === 'dark' ? 'dark' : undefined,
-            }}
-          />
-        )
+        // The first CTA and second CTA should always
+        // have the `primary` and `secondary` variations respectively.
+        // This const is confusingly named because `Button` and `StandaloneLink`
+        // use different prop names to trigger variations (`variant` and `theme`).
```
nandereck, line 25: TODO: delete 2nd part of this comment.

nandereck:
> Update props.js to fix `brand` -> `theme`. This was my mistake when initially building this out. and I don't think variant is available on the ctas, its type right?

Updated the docs so should be up-to-date with my change.

.changeset/tidy-emus-hammer.md
```diff
@@ -0,0 +1,7 @@
+---
+'@hashicorp/react-actions': major
+'@hashicorp/react-intro': major
+'@hashicorp/react-next-steps': major
+---
```
nandereck, line 5: @alexcarpenter - just want to triple check that we want all packages to be major updates since this introduces a breaking change, right?

alexcarpenter: I think am thinking so. Does that sound right @dstaley?

dstaley: If I'm understanding correctly, only `react-actions` should have a changeset entry. You'd remove the changes to the rests for `intro` and `next-steps`, and update them to the new major version in a subsequent PR. Also, I believe you might need to merge in changes from `main` as it looks like some of your `package-lock.json` changes are from other PRs?

nandereck: Ahhh, that makes sense. ⚡  So, this is what I need to do:

- Update changeset such that there's only a major bump for `actions`
- Revert changes to `intro` and `next-steps`
- Merge in / publish
- Create separate PRs for `intro` and `next-steps` changes that include bumping to the new `actions` version ☝️

dstaley: Yup! That's the general procedure for releasing a change that has to be fixed/upgraded to in dependent packages.

nandereck:
Paired with @dstaley and decided that all changes should be in this PR (versus separate). Consequently, the changeset here is minor bumps to all three packages. I think this is ready again for review.

alexcarpenter (approved):
LGTM! Nice work. 👏🏼

PR #759, opened 2022-10-12
Make tile `background-color` white to support different background colors

nandereck:
🎟️ [Asana Task](https://app.asana.com/0/1202802428287999/1202891942449714/f)

---

## Description
As part of Use Case QA, design requested that the `NextSteps` component support a gray background. While it's unclear at the moment if we want to add this functionality to the `NextSteps` component more broadly, we do want to be able to do this specifically for Use Cases. This is easily achieved by wrapping `NextSteps` in a div and setting the `background-color`, however, it was discovered that the non-primary tiles have a transparent background. This results in the tile and `NextSteps` components being the same color which is not desireable.

This PR sets the `background-color` for non-primary tiles to white so that changes to the wrapping element's `background-color` don't also change the `background-color` of the tile.

## Validation Steps
- Navigate to the [`Next Steps` component in the preview](https://react-components-git-nels-next-stepsadd-backgr-0355cd-hashicorp.vercel.app/components/nextsteps).
- Looking at the light and dark themes, validate that there are no visual regressions

### PR Checklist 🚀

### DIFF
--- a/.changeset/hungry-bears-tell.md
+++ b/.changeset/hungry-bears-tell.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-next-steps': minor
+---
+
+Set non-primary tile background color to white
--- a/packages/next-steps/style.module.css
+++ b/packages/next-steps/style.module.css
@@ -85,8 +85,8 @@
 }
 
 .tile {
-  --background-color: transparent;
-  --border-color: transparent;
+  --background-color: var(--wpl-neutral-0);
+  --border-color: --wpl-neutral-0;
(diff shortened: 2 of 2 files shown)


nandereck:
@alexcarpenter - do you know if there's a specific reason why we set the `background-color` to transparent? Is it because we always want the tile background to match the background of the component? 

Reason I'm asking is because the Use Case spec calls for a gray background and white tiles. My change would help achieve this but not sure if I'm missing anything.

alexcarpenter (approved):
:shipit:

PR #767, opened 2022-10-19
feat: `<ProductsUsed />` component

nandereck:
🎟️ [Asana Task](https://app.asana.com/0/1202802428287999/1203075725709005/f)
🔍 [Preview Link](https://react-components-git-nels-products-usedcreate-f74ae8-hashicorp.vercel.app/components/productsused)
🎨 [Figma](https://www.figma.com/file/VvpEQaWhKQExx9QTWRyayd/Patterns?node-id=677%3A4054)

---

## Description

Adds new `<ProductsUsed />` component:
- Bumps `mktg-global-styles` so as to utilize spacing tokens
- Adds docs, props, types and component files for `ProductsUsed`

## Validation Steps
- Pull down branch and run `npm run test:watch`
- Verify no failed tests
- Navigate to the [component in Preview](https://react-components-git-nels-products-usedcreate-f74ae8-hashicorp.vercel.app/components/productsused)
- Verify component [matches the spec](https://www.figma.com/file/VvpEQaWhKQExx9QTWRyayd/Patterns?node-id=677%3A4054)

### PR Checklist 🚀

### DIFF
--- a/packages/products-used/index.tsx
+++ b/packages/products-used/index.tsx
@@ -9,7 +9,7 @@ const ProductsUsed = ({
   const eyebrowText = 'Products used'
 
   return (
-    <div className={classNames([s.root, s[appearance]])}>
+    <div className={classNames([s.root, s[appearance]])} data-testid="root">
       <p className={classNames([s.eyebrow, s[appearance]])}>{eyebrowText}</p>
       <ul className={s.productList}>
         {products.map((product, index) => {
@@ -50,7 +50,7 @@ const ProductsUsed = ({
 
 const ConditionalLink = ({ href, children }) => {
   return href ? (
-    <a href={href} className={s.isVerticallyCentered}>
+    <a href={href} className={s.isVerticallyCentered} data-testid={'anchorEl'}>
       {children}
     </a>
   ) : (
--- a/packages/products-used/index.tsx
+++ b/packages/products-used/index.tsx
@@ -27,25 +27,33 @@ const ProductsUsed = ({
 }: ProductsUsedProps) => {
   const eyebrowText = 'Products used'
 
+  // Products with links have different spacing between each other (8px) than
+  // products without links (16px). This spacing is set on the `ul` using the `gap` property.
+  // This logic determines whether or not products have links so as to conditionally change
+  // the spacing.
+  const productsWithLinks = products.find((product) =>
+    Object.keys(product).includes('href')
+  )
+  const productsHaveLinks = !!productsWithLinks
+
   return (
     <div className={classNames([s.root, s[appearance]])} data-testid="root">
       <p className={classNames([s.eyebrow, s[appearance]])}>{eyebrowText}</p>
-      <ul className={s.productList}>
+      <ul
+        className={classNames([s.productList], {
+          [s.isSmallerGap]: productsHaveLinks,
+        })}
+      >
         {products.map((product, index) => {
-          const isLink = !!product.href
-
           return (
             <li
               // eslint-disable-next-line react/no-array-index-key
               key={index}
-              className={classNames([
-                s.product,
-                s.isVerticallyCentered,
-                s[appearance],
-                { [s.isLink]: isLink },
-              ])}
             >
-              <ConditionalLink href={product.href}>
+              <ConditionalLink
+                className={classNames([s.product, s[appearance]])}
+                href={product.href}
+              >
                 <span className={s.productIcon}>
                   {ProductIconMap[product.name]}
                 </span>
@@ -61,13 +69,13 @@ const ProductsUsed = ({
   )
 }
 
-const ConditionalLink = ({ href, children }) => {
+const ConditionalLink = ({ href, children, className }) => {
   return href ? (
-    <a href={href} className={s.isVerticallyCentered} data-testid={'anchorEl'}>
+    <a href={href} className={className} data-testid={'anchorEl'}>
       {children}
     </a>
   ) : (
-    children
+    <div className={className}>{children}</div>
   )
 }
 
--- a/packages/products-used/index.tsx
+++ b/packages/products-used/index.tsx
@@ -31,10 +31,7 @@ const ProductsUsed = ({
   // products without links (16px). This spacing is set on the `ul` using the `gap` property.
   // This logic determines whether or not products have links so as to conditionally change
   // the spacing.
-  const productsWithLinks = products.find((product) =>
-    Object.keys(product).includes('href')
-  )
-  const productsHaveLinks = !!productsWithLinks
+  const productsHaveLinks = products[0]?.hasOwnProperty('href')
 
   return (
     <div className={classNames([s.root, s[appearance]])} data-testid="root">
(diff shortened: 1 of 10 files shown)


packages/products-used/style.module.css
```diff
@@ -0,0 +1,109 @@
...
+  & .isVerticallyCentered {
+    display: flex;
+    align-items: center;
+  }
+}
+
+.product {
+  &:not(:first-child) {
+    margin-left: var(--wpl-spacing-03);
+  }
```
alexcarpenter, line 31: Can we rely on gap vs margins?

packages/products-used/types.ts
```diff
@@ -0,0 +1,26 @@
...
+   * Array of products or products with links. Only supports one or the other, not a mix of the two.
+   */
+  products: Array<Product> | Array<ProductLink>
+}
+
+interface Product {
+  /**
+   * Renders a product name
+   */
+  name: Products
```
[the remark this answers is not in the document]
nandereck: Ah, yes yes.

packages/products-used/style.module.css
```diff
@@ -0,0 +1,109 @@
...
+  color: var(--wpl-neutral-500);
+
+  &.dark {
+    color: var(--wpl-neutral-400);
+  }
+}
+
+.productList {
+  display: flex;
+  align-items: center;
```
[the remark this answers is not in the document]
nandereck: Ooh, yup!

packages/products-used/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const eyebrowText = 'Products used'
+
+  return (
+    <div className={classNames([s.root, s[appearance]])} data-testid="root">
+      <p className={classNames([s.eyebrow, s[appearance]])}>{eyebrowText}</p>
+      <ul className={s.productList}>
+        {products.map((product, index) => {
+          const isLink = !!product.href
+          const iconStyles = {
+            '--icon': `url("https://www.hashicorp.com/img/flight-icons/${product.name}-color-24.svg")`,
```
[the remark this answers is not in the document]
nandereck: Did _not_ think of that. Fixed!

packages/products-used/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+          return (
+            <li
+              // eslint-disable-next-line react/no-array-index-key
+              key={index}
+              className={classNames([
+                s.product,
+                s.isVerticallyCentered,
+                s[appearance],
+                { [s.isLink]: isLink },
+              ])}
```
alexcarpenter, line 46: I might be missing it, but are all these classes being applied here needed/being used in some fashion?

nandereck: Stand by... this is going to change when I fix the clickability issue and replace margin with gap.

alexcarpenter: we can detect link usage using CSS so not sure we need to use JS to supply the isLink class.

```scss
// Apply inner class to both link and non-link versions.
.inner {
  &[href] {
    // If we have an href, were in a link. Apply link styling.
  }
}
```

packages/products-used/style.module.css
nandereck: One thing to flag is that products with links have different spacing between each other than products without links (8px and 16px respectively). Using `gap` would require that the parent that renders product items knows whether or not the items are links or non-links.

Not an issue per se but using gap does require some extra logic to expose this to the parent.

alexcarpenter: Ahh I missed that so we do need the includesLink logic to be applied to the ul in some fashion 👍🏼. I think we'd still prefer this to be set via gap verses margin.

packages/products-used/style.module.css
```diff
@@ -0,0 +1,104 @@
...
+        border-color: var(--wpl-neutral-500);
+      }
+
+      &:disabled {
+        border-color: var(--wpl-neutral-900);
+
+        &:hover {
+          background-color: var(--wpl-neutral-900);
+        }
+      }
```
alexcarpenter, line 85: I know its include in figma, but there is no such thing as a disabled link. So I think it's safe to remove this.

packages/products-used/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+  appearance = 'light',
+  products,
+}: ProductsUsedProps) => {
+  const eyebrowText = 'Products used'
+
+  // Products with links have different spacing between each other (8px) than
+  // products without links (16px). This spacing is set on the `ul` using the `gap` property.
+  // This logic determines whether or not products have links so as to conditionally change
+  // the spacing.
+  const productsHaveLinks = products[0]?.hasOwnProperty('href')
```
nandereck, line 34: @alexcarpenter - what're your thoughts on this logic? I don't love it.

alexcarpenter: This works as expected. Another alternative might be `products.some((x) => x.href)` but I think since at the type level we are enforcing them all having hrefs or all not having hrefs, what you have is the way to go.

packages/products-used/style.module.css
```diff
@@ -0,0 +1,84 @@
...
+  }
+}
+
+.productList {
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
+  padding: 0;
+  list-style-type: none;
+  gap: var(--wpl-spacing-03);
```
alexcarpenter, line 23: Looks like we have a consistent row-gap of 8px and only the column-gap changes from 8px to 16px.

packages/products-used/style.module.css
```diff
@@ -0,0 +1,84 @@
...
+      &:hover {
+        background-color: var(--wpl-neutral-850);
+        border-color: var(--wpl-neutral-500);
+      }
+
+      &:active {
+        background-color: var(--wpl-neutral-800);
+        border-color: var(--wpl-neutral-500);
+      }
+    }
```
alexcarpenter, line 66: Is there a dark focus state we need to account for?

packages/products-used/style.module.css
```diff
@@ -0,0 +1,84 @@
...
+}
+
+.product {
+  display: flex;
+  align-items: center;
+
+  &[href] {
+    padding: var(--wpl-spacing-01) var(--wpl-spacing-02);
+    border-radius: calc(var(--wpl-spacing-01) / 2);
+    border: 1.5px solid var(--wpl-neutral-100);
```
alexcarpenter, line 37: Always tend to be hesitant to use half pixels as values. What do yo think?

packages/products-used/style.module.css
nandereck: Ah, good catch. Updated.

packages/products-used/style.module.css
nandereck: Oops - it's the same for light and dark and I forgot to add it for dark. 👍

packages/products-used/style.module.css
nandereck: Same - bumped to 2px.

nandereck:
@alexcarpenter - I've addressed your feedback. Mind giving this another look? 🙌

alexcarpenter (approved):
:shipit:

PR #736, opened 2022-10-03
feat(code-block): support multiple single-line cmd copy

zchsh:
🎟️ [Asana Task][task]
🔍 [Preview Link][preview]

---

Support copy-button copying of shell-session snippets with multiple single-line commands. 

Support select-copy-paste copying of shell-session snippets by setting `user-select: none` CSS on `.shell-symbol` tokens.

> **Note**: pre-releases has been tested in hashicorp/dev-portal#1149.

### PR Checklist 🚀

Items in this checklist may not may not apply to your PR, but please consider each item carefully.

- [x] Add Asana and Preview links above.
- [x] Conduct thorough self-review.
- [x] Add or update tests as appropriate.
- [x] Conduct reasonable cross browser testing for both compatibility and responsive behavior (We have a [Sauce Labs](https://app.saucelabs.com/) account for this, if you don't have access, just ask!).
- [x] Conduct reasonable accessibility review (use the [WAS](https://accessible.org/Web-Accessibility-Standards-WAS-2.pdf) as a guide or an [axe browser plugin](https://www.deque.com/axe/) until we establish more formal checks).
- [x] Identify (in the description above) and document (add Asana tasks on [this board](https://app.asana.com/0/1100423001970639/list)) any technical debt that you're aware of, but are not addressing as part of this PR.

[task]: https://app.asana.com/0/1202097197789424/1203067508838442/f
[preview]: https://react-components-git-zscode-block-multi-command-copy-hashicorp.vercel.app/components/codeblock

PR #768, opened 2022-10-19
fix(code-block): fix language-undefined in MDX CodeTabs

zchsh:
🎟️ [Asana Task](https://app.asana.com/0/1202097197789424/1203204105610771/f)
🔍 [Preview Link](https://react-components-git-zsfix-code-tabs-language-hashicorp.vercel.app)

---

Fixes a bug where `shell-session` snippets within `CodeTabs` would copy correctly via the `Copy` button, but would still include the `$` character when selected and copied by clicking and dragging.

## Details

Removes a `className` attribute that passed a non-existent className, which also dropped any incoming classNames. This was originally intended to prevent the possibility of custom styling via MDX, but has the side-effect of dropping the `language-<language>` class. That language class was unused for the most part (highlighting happens on the server!), but has become important for full functionality since #736 shipped.

This approach has been proven out in https://github.com/hashicorp/dev-portal/pull/1268. Screenshots below are from that PR, demonstrating that with this fix, we go from the issue-causing `language-undefined` to `language-shell-session`.

### Before

before

### After

after

### PR Checklist 🚀

### DIFF
--- a/.changeset/itchy-flowers-whisper.md
+++ b/.changeset/itchy-flowers-whisper.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-code-block': patch
+---
+
+Fixes issue where code-tabs in MDX would render blocks with a language-undefined className, even if the className was in fact defined.
--- a/package-lock.json
+++ b/package-lock.json
@@ -35335,7 +35335,7 @@
     },
     "packages/actions": {
       "name": "@hashicorp/react-actions",
-      "version": "0.3.0",
+      "version": "0.4.0",
       "license": "MPL-2.0",
       "dependencies": {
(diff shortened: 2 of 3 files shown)


BRKalow (approved):
Good catch! Do you think there's any reason why we would want to retain the ignoring of other passed `className` values?

zchsh:
@BRKalow I think my previous reasoning was to try to preemptively prevent any custom styling in MDX contexts. But in retrospect I don't think that made all that much sense, custom styling would probably happen through the `style` prop anyways? 🤔 So I don't see any strong reason to retain the "ignore incoming `className`" behaviour.

BRKalow:
Makes sense, thanks for the insight!

PR #773, opened 2022-10-21
chore(NextSteps): update cta styling and animation

alexcarpenter:
🔍 [Preview Link](https://react-components-git-acnext-steps-animation-hashicorp.vercel.app/components/nextsteps)

---

## Description

Updates NextSteps Cta styling and animation to match what we are doing in StandaloneLink.

https://user-images.githubusercontent.com/825855/197267885-e6ba9796-5203-4237-8852-c52d00b2f76c.mp4

### PR Checklist 🚀

### DIFF
--- a/packages/next-steps/index.tsx
+++ b/packages/next-steps/index.tsx
@@ -61,7 +61,31 @@ export default function NextSteps({
                       <p className={s.tileDescription}>{step.description}</p>
                     ) : null}
                     <span className={s.tileCta}>
-                      {step.cta.title} <IconArrowRight16 />
+                      <span className={s.tileCtaLabel}>{step.cta.title}</span>
+                      <span className={s.tileCtaIcon}>
+                        <svg
+                          className={s.arrow}
+                          width="9"
+                          height="10"
+                          viewBox="0 0 9 10"
+                          fill="none"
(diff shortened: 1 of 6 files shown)


packages/next-steps/index.tsx
```diff
@@ -61,7 +61,31 @@ export default function NextSteps({
                       <p className={s.tileDescription}>{step.description}</p>
                     ) : null}
                     <span className={s.tileCta}>
-                      {step.cta.title} <IconArrowRight16 />
+                      <span className={s.tileCtaLabel}>{step.cta.title}</span>
```
nandereck, line 64: 💭 Non-blocking: looks like we're copying this over from `StandaloneLink`. Any thoughts on abstracting this or parts of this out somewhere?

nandereck (approved):
LGTM 🎯

packages/next-steps/index.tsx
alexcarpenter: The tricky part when thinking about abstracting this is the parent element that is triggering the hover animation. In StandaloneLink it is the link, in NextSteps it's the tile. Not to say its not possible but it felt ok to copy paste this for now.

nandereck: _Oh interestin'_. That makes sense!

PR #776, opened 2022-10-24
[Assembly] New variants for `<Card />`

EnMod:
🔍 [Preview Link](https://react-components-git-asm-new-card-variants-hashicorp.vercel.app/components/card)

---

## Description

Assembly collecting changes to `<Card />` to support two new variants: `<PersonCard />` and `<EventCard />`.

### PR Checklist 🚀

### DIFF
--- a/packages/card/style.module.css
+++ b/packages/card/style.module.css
@@ -4,20 +4,12 @@
   --secondary-text-color: var(--wpl-neutral-500);
   --background-color: var(--wpl-neutral-0);
 
-  border: 1px solid var(--border-color);
-  display: flex;
-  flex-direction: column;
-  flex-grow: 1;
-  border-radius: 8px;
-  overflow: hidden;
-  position: relative;
-  background-color: var(--background-color);
-  transition: 0.2s ease;
-  transition-property: transform, border-color, box-shadow, background-color;
-  box-shadow: 0 16px 24px -16px #cbcbcd33, 0 8px 12px -8px #cbcbcd66;
-
   &:hover {
     --border-color: var(--wpl-neutral-300);
+
+    & .contentContainer {
+      transform: translateY(-2px);
+    }
 
     & .cta svg {
       transform: translateX(2px);
@@ -37,8 +29,9 @@
     box-shadow: 0 16px 24px -16px #cbcbcd33, 0 8px 12px -8px #cbcbcd66;
 
     &:hover {
-      box-shadow: 0 16px 16px -10px #656a7633, 0 2px 3px 0 #656a7626;
-      transform: translateY(-2px);
+      & .contentContainer {
+        box-shadow: 0 16px 16px -10px #656a7633, 0 2px 3px 0 #656a7626;
+      }
 
       & .cta {
         color: var(--border-color);
@@ -56,6 +49,19 @@
       --background-color: var(--wpl-neutral-700);
     }
   }
+}
+
+.contentContainer {
+  border: 1px solid var(--border-color);
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
+  border-radius: 8px;
+  overflow: hidden;
+  position: relative;
+  background-color: var(--background-color);
+  transition: 0.2s ease;
+  transition-property: transform, box-shadow, border-color, background-color;
 }
 
 .link {
--- a/packages/card/style.module.css
+++ b/packages/card/style.module.css
@@ -19,9 +19,6 @@
   &:hover {
     --border-color: var(--wpl-neutral-300);
 
-    box-shadow: 0 16px 16px -10px #656a7633, 0 2px 3px 0 #656a7626;
-    transform: translateY(-2px);
-
     & .cta svg {
       transform: translateX(2px);
     }
@@ -37,7 +34,12 @@
   }
 
   &.light {
+    box-shadow: 0 16px 24px -16px #cbcbcd33, 0 8px 12px -8px #cbcbcd66;
+
     &:hover {
+      box-shadow: 0 16px 16px -10px #656a7633, 0 2px 3px 0 #656a7626;
+      transform: translateY(-2px);
+
       & .cta {
         color: var(--border-color);
       }
--- a/packages/card/style.module.css
+++ b/packages/card/style.module.css
@@ -76,7 +76,9 @@
 }
 
 .content {
-  display: grid;
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
   row-gap: 16px;
   padding: 24px 24px 0 24px;
 }
--- a/packages/card/style.module.css
+++ b/packages/card/style.module.css
@@ -4,6 +4,10 @@
   --secondary-text-color: var(--wpl-neutral-500);
   --background-color: var(--wpl-neutral-0);
 
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
+
   &:hover {
     --border-color: var(--wpl-neutral-300);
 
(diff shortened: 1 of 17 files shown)


packages/card/style.module.css
```diff
@@ -13,11 +13,15 @@
...
   transition: 0.2s ease;
-  transition-property: border-color, background-color;
+  transition-property: transform, border-color, box-shadow, background-color;
+  box-shadow: 0 16px 24px -16px #cbcbcd33, 0 8px 12px -8px #cbcbcd66;
 
   &:hover {
     --border-color: var(--wpl-neutral-300);
 
+    box-shadow: 0 16px 16px -10px #656a7633, 0 2px 3px 0 #656a7626;
+    transform: translateY(-2px);
```
alexcarpenter, line 23: Doom flicker https://www.joshwcomeau.com/animation/css-transitions/#doom-flicker


https://user-images.githubusercontent.com/825855/198085740-20c7c333-5aeb-48e6-9211-dbd295cc626b.mp4

packages/card/style.module.css
```diff
@@ -13,11 +13,15 @@
   position: relative;
   background-color: var(--background-color);
   transition: 0.2s ease;
-  transition-property: border-color, background-color;
+  transition-property: transform, border-color, box-shadow, background-color;
+  box-shadow: 0 16px 24px -16px #cbcbcd33, 0 8px 12px -8px #cbcbcd66;
```
alexcarpenter, line 17: I think the intent based on Rays comment, is that this box-shadow is only applied in light appearance.

EnMod: Ah, yes that's a good point. Addressed in the latest commits

packages/card/style.module.css
EnMod: 😱 Good catch! Thanks for that link as well, it was useful in understanding the issue in greater detail :+1: 

Addressed here:
https://github.com/hashicorp/react-components/pull/776/commits/6b507ce9f1784b2c0acfca9c142894af289a336c

packages/card/style.module.css
```diff
@@ -4,20 +4,13 @@
   --secondary-text-color: var(--wpl-neutral-500);
   --background-color: var(--wpl-neutral-0);
 
-  border: 1px solid var(--border-color);
-  display: flex;
-  flex-direction: column;
-  flex-grow: 1;
```
alexcarpenter, line 10: I think this is still needed at the root level to ensure the card grows the the height of containing element. See the last card is not the same height.

CleanShot 2022-10-26 at 13 23 55@2x

EnMod: ❗ Yep, absolutely needs to stay there as well :+1: 

Addressed here:
https://github.com/hashicorp/react-components/pull/776/commits/dd5925b910769a3050a322764b0973e7af208aec

alexcarpenter:
Lets update to use [`<ExpandableArrow />`](https://react-components.vercel.app/components/expandablearrow) vs custom implementation. Ref: https://github.com/hashicorp/react-components/pull/806

EnMod:
@alexcarpenter Implemented `<ExpandableArrow />` in the latest commit:
https://github.com/hashicorp/react-components/pull/776/commits/42c52885d8ea727f4e14d88d52a9dd16d17994ef

The `Card` typing could use some work now though 😅

EnMod:
Rebased on `main` to resolve conflicts

packages/card/card.tsx
```diff
@@ -0,0 +1,151 @@
...
+  ProductBadgesProps,
+  DescriptionProps,
+} from './types'
+import s from './style.module.css'
+
+function Card(props: CardProps) {
+  const {
+    appearance = 'light',
+    withArrow = true,
+    hoverRef,
```
alexcarpenter, line 21: Should we be forwarding refs here? https://reactjs.org/docs/forwarding-refs.html

EnMod: Ah, yes we probably should be! It seems this component is a good use case for it.

EnMod: Addressed here, need to sort out TS typing now that the component is using `forwardRef` 🤔 
https://github.com/hashicorp/react-components/pull/776/commits/94ec2dd71a975df57df5767df3f77619274cc9be

packages/card/card.tsx
```diff
@@ -0,0 +1,165 @@
...
+    CardProps & React.RefAttributes<HTMLDivElement>
+  > {
+  Thumbnail?: $TSFixMe
+  Meta?: $TSFixMe
+  Content?: $TSFixMe
+  Heading?: $TSFixMe
+  Description?: $TSFixMe
+  ProductBadges?: $TSFixMe
+}
+
```
EnMod, line 30: 🤔 Adding typing for this card as a compound component has proven tricky.  This is the direction I've seen as a working approach, but obviously without the `$TSFixMe` placeholders 😅 

Just added those to get working builds again and put the approach up for discussion!

EnMod: Update: I've refactored the component to use a different approach.  

Instead of forwarding a `ref` to the parent to use with hover state logic via `forwardRef`, I instead implemented a `hoverHandler` prop and moved the `useHover()` hook into the `Card` component proper. I added a `useEffect()` to `Card` that fires when `isHovered` changes, and executes `hoverHandler` if that's passed into the `Card`.  

This does introduce unnecessary `useEffect()` calls if there's no `hoverHandler` passed to `Card`, but this does sidestep the TS typing issues here (and makes the code a bit more flexible for consumers looking to execute logic on `Card` hover).  I'm looking into how to prevent the unnecessary `useEffect()` calls on no `hoverHandler` passed 🤔

packages/card/style.module.css
```diff
@@ -1,23 +1,20 @@
...
   --primary-text-color: var(--wpl-neutral-700);
-  --secondary-text-color: var(--wpl-neutral-500);
+  --secondary-text-color: #75757a;
   --background-color: var(--wpl-neutral-0);
 
-  border: 1px solid var(--border-color);
   display: flex;
   flex-direction: column;
   flex-grow: 1;
-  border-radius: 8px;
```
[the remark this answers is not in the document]
EnMod: Ah yes, that may have been removed somehow in the rebase. Adressed here:
https://github.com/hashicorp/react-components/pull/776/commits/e2df01db5c4d1d2a988330e5c0a78a37caa29d89

packages/card/style.module.css
```diff
@@ -1,23 +1,21 @@
...
-  border: 1px solid var(--border-color);
   display: flex;
   flex-direction: column;
   flex-grow: 1;
   border-radius: 8px;
-  overflow: hidden;
-  position: relative;
-  background-color: var(--background-color);
-  transition: 0.2s ease;
-  transition-property: border-color, background-color;
```
pbortnick, line 16: Similar to the border-radius comment, I think removing these styles here causes extra spacing on hover + focus. Kind of nitpicky - not sure if we are already okay with this 
Screen Shot 2022-11-11 at 1 10 10 PM

EnMod: Hmm...still getting the same issue when adding these lines back in 🤔 

6T6Bs7VeBW

alexcarpenter: Likely need to add the outline styles to the a tag or the contentContainer since that is what translates on hover.

EnMod: Moved the `&:focus-within` styles to `.contentContainer`, and the outline moves successfully 👍
https://github.com/hashicorp/react-components/pull/776/commits/519a4e94d2b218d390ab72d1e978794e58c5f7f2

pbortnick: Oh yeah looks really good now 😄

pbortnick (approved):
✨

PR #777, opened 2022-10-24
[WIP] feat(Card): `<PersonCard />` variant

EnMod:
🔍 [Preview Link](https://react-components-git-nqmcv-personcard-variant-hashicorp.vercel.app/components/card)

---

## Description

This PR creates a `<PersonCard />` variant of the `<Card />` component.

### PR Checklist 🚀


packages/card/index.tsx
```diff
@@ -132,10 +132,52 @@ function Description({ children }: DescriptionProps) {
...
+}
+
 Card.Thumbnail = Thumbnail
 Card.Meta = Meta
 Card.Content = Content
 Card.Heading = Heading
 Card.Description = Description
+Card.PersonCard = PersonCard
 
 export default Card
```
alexcarpenter, line 183:
```suggestion

export default Card
export { PersonCard }
```

This would actually be a named export vs adding to the compound component.

EnMod: Ah, it sure would! Good catch :+1:

packages/card/index.tsx
```diff
@@ -132,10 +132,52 @@ function Description({ children }: DescriptionProps) {
...
+  productBadges,
+}: PersonCardProps) {
+  return (
+    // TODO Determine how to make these two props optional for PersonCard
+    // TODO Possibly make their requirement dependent on whether children are passed or not? May need a type guard function...
+    <Card heading={name} link={link} appearance={appearance}>
+      <Thumbnail {...thumbnail} />
+      <Content>
+        <Heading>{name}</Heading>
+        {/* ? TODO Possibly extract location to a component for text of this style? */}
```
alexcarpenter, line 165: Feels fine to leave as is for now.

EnMod:
Not adding a changeset or tests for this variant here just yet:
- I'm thinking it might be best to tackle tests in a single PR after both variants merge to assembly. That way both variants' APIs can stabilize and there will be a more solid base to test from.
- A changeset update might be best done in assembly before merge, after all changes are ready to merge to `main`. That way the final changeset can capture the changes without any conflicts arising from merging two different changesets to a single branch.

package-lock.json
```diff
@@ -36478,7 +36478,7 @@
     },
     "packages/next-steps": {
       "name": "@hashicorp/react-next-steps",
-      "version": "0.3.0",
+      "version": "0.4.0",
```
alexcarpenter, line 36481: Unintended change?

packages/card/person-card/style.module.css
```diff
@@ -0,0 +1,35 @@
+.location {
+  composes: g-type-body-small-x-strong from global;
+  margin: 0;
+  color: var(--secondary-text-color);
```
alexcarpenter, line 4: hmm, does this color get applied correctly even though `--secondary-text-color` is not defined in this file?

package-lock.json
EnMod: Hmmm...it occurred after I ran `npm i` to get installed deps up to date 🤔 I can revert before merge though

packages/card/person-card/style.module.css
EnMod: Yep; verified that `--secondary-text-color` is successfully cascading down from `<Card />`:

Slack_LlUdi8ihpw

package-lock.json
alexcarpenter: ah ok, its in the lock file so I guess that should be ok!

alexcarpenter (approved):

PR #780, opened 2022-10-25
feat(Card): `<EventCard />` variant

EnMod:
🔍 [Preview Link](https://react-components-git-nqncv-eventcard-variant-hashicorp.vercel.app/components/card)

---

## Description

### PR Checklist 🚀

### DIFF
--- a/packages/card/event-card/index.tsx
+++ b/packages/card/event-card/index.tsx
@@ -1,5 +1,6 @@
 import Card from '../card'
 import { CardProps, ProductBadgesProps } from '../types'
+import s from './style.module.css'
 
 interface EventCardProps {
   appearance?: CardProps['appearance']
@@ -22,13 +23,44 @@ export function EventCard({
 
   return (
     <Card heading={title} link={link} appearance={appearance}>
-      <Card.Content>
-        <Card.Meta items={[eventDate, eventType]}></Card.Meta>
-        <Card.Heading>{title}</Card.Heading>
-        {productBadges && productBadges?.length > 0 ? (
-          <Card.ProductBadges badges={productBadges} appearance={appearance} />
-        ) : null}
-      </Card.Content>
+      <div className={s.eventCardInner}>
+        <Card.Content>
+          <Card.Meta items={[eventDate, eventType]}></Card.Meta>
+          <Card.Heading>{title}</Card.Heading>
+          {productBadges && productBadges?.length > 0 ? (
+            <Card.ProductBadges
+              badges={productBadges}
+              appearance={appearance}
+            />
+          ) : null}
+          <span className={s.cta}>
+            <span className={s.ctaLabel}>{ctaText}</span>
+            <span className={s.ctaIcon}>
+              <svg
+                className={s.ctaIconArrow}
+                width="9"
+                height="10"
+                viewBox="0 0 9 10"
+                fill="none"
+                xmlns="http://www.w3.org/2000/svg"
+                focusable={false}
+              >
+                <path
+                  d="M4 1L8 5L4 9"
+                  stroke="currentColor"
+                  strokeWidth={1.8}
+                />
+                <path
+                  className={s.ctaIconLine}
+                  d="M8 5H0"
+                  stroke="currentColor"
+                  strokeWidth={1.8}
+                />
+              </svg>
+            </span>
+          </span>
+        </Card.Content>
+      </div>
     </Card>
   )
 }
--- a/packages/card/event-card/index.tsx
+++ b/packages/card/event-card/index.tsx
@@ -21,8 +21,6 @@ export function EventCard({
   const ctaText = `Register for ${eventType.toLowerCase()}`
 
   return (
-    // TODO Determine how to make these two props optional for EventCard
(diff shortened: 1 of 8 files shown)


packages/card/event-card/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <Card.Content>
+          <Card.Meta items={[eventDate, eventType]}></Card.Meta>
+          <Card.Heading>{title}</Card.Heading>
+          {productBadges && productBadges?.length > 0 ? (
+            <Card.ProductBadges
+              badges={productBadges}
+              appearance={appearance}
+            />
+          ) : null}
+          <span className={s.pseudoCta}>
```
alexcarpenter, line 36: Assuming this pseudoCta should be anchored to the bottom of the card. Maybe something to confirm with design.

CleanShot 2022-10-25 at 11 32 12@2x

alexcarpenter: Might have to reconsider withArrow functionality to make this happen 🤔

packages/card/event-card/style.module.css
```diff
@@ -0,0 +1,48 @@
...
+.pseudoCtaIconArrow {
+  --translate-x: -4px;
+
+  transform: translateX(var(--translate-x));
+  transition-duration: var(--transition-duration);
+  transition-timing-function: var(--transition-timing-function);
+  transition-property: transform;
+  color: inherit;
+
+  @nest [data-testid='wpl-card']:hover & {
```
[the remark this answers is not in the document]
EnMod: Agreed, and I _would_ target the root class, but the root class in this case is in `<Card />` so I can't target it directly from `<EventCard />`'s CSS (right?). I think I could perhaps wrap `<EventCard />` in a root div, then target that?

packages/card/event-card/index.tsx
EnMod: I still need to bring over the `withArrow` prop/logic from #777, once I do the card will match spec and I'll commit your other suggestion (thanks for that)

packages/card/event-card/style.module.css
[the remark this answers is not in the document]
alexcarpenter: Makes me think we need to add className prop to base card.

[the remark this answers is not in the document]
EnMod: Definitely agree there, otherwise imagine the other workarounds we'd need to deal with for similar behavior in the future 😱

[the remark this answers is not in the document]
alexcarpenter: or maybe we add a global className to the card that we can reference?

```tsx
<div className={classNames('wpl-card', s.card, s[appearance])} data-testid="wpl-card">
```

```css
@nest :global(.wpl-card):hover & {...}
```

Not validated that this works but just another idea.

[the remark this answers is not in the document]
EnMod: Hmm...I do like the global class option vs adding support for `className`. Using a global class will discourage making one-off modifications to cards' root CSS, which we probably want to avoid doing if we can help it, for styling/CSS consistency.

packages/card/event-card/index.tsx
alexcarpenter: Right, but the pseudoCta wont be able to be bottom aligned since the cta div is still in the document and taking up the space at the bottom of the card.

packages/card/event-card/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+  eventDate,
+  eventType,
+  title,
+  productBadges,
+}: EventCardProps) {
+  const ctaText = `Register for ${eventType.toLowerCase()}`
+
+  return (
+    <Card heading={title} link={link} appearance={appearance} withArrow={false}>
+      <div className={s.eventCardInner}>
```
alexcarpenter, line 26: Can we remove this wrapping div? No CSS is being applied here and removes the ability to use flexbox auto margins on children.

EnMod: Yep that's an artifact of some earlier attempts at making a wrapper to get hover state from. 🔪 shortly!

packages/card/event-card/index.tsx
EnMod: Hmm...it looks like it's flush with `<Content />`'s padding of `24px`, now that I've updated the card with `withArrow` 🤔 

chrome_1L9nFpnMMr

EnMod: We discussed this and I hadn't tested with different event title content, so I never noticed how there was no guaranteed alignment of the pseudo-CTA with the bottom of the card. 

This issue was addressed in this commit:
https://github.com/hashicorp/react-components/pull/780/commits/1627dfb1758bd318d41fa1d47378f30b3e072595

packages/card/event-card/style.module.css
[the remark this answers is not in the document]
EnMod: Implemented the global class option in the latest two commits, works well I think!
https://github.com/hashicorp/react-components/pull/780/commits/5a155509289be30cf887cf15f69d5ae5c335ac4e
https://github.com/hashicorp/react-components/pull/780/commits/f3c76913e8bf3be35e607a6dc43de63b24ce01bb

PR #782, opened 2022-10-26
feat(Card): Add new tests for variants

EnMod:
🔍 [Preview Link](https://react-components-git-{branch-slug}-hashicorp.vercel.app/components/card)

---

## Description

This PR adds tests for the new `<Card />` component variants.

### PR Checklist 🚀

### DIFF
--- a/packages/card/event-card/index.test.js
+++ b/packages/card/event-card/index.test.js
@@ -1,0 +1,37 @@
+import { render, screen } from '@testing-library/react'
+import { EventCard } from '.'
+
+const defaultProps = {
+  link: 'https://hashicorp.com',
+  title: 'Example title',
+  eventDate: 'August 15, 2022',
+  eventType: 'Category',
+  thumbnail: {
+    src: 'https://www.datocms-assets.com/19447/1648680074-screenshot-2022-03-31-at-00-40-47.png',
+    alt: 'HashiConf Europe 2022 Recap',
+  },
+}
+
+describe('<EventCard />', () => {
+  // Prop tests
+  it('should render the provided title prop', () => {
+    render(<EventCard {...defaultProps} />)
+
+    const title = screen.getByText(defaultProps.title)
+    expect(title.tagName).toBe('H2')
+  })
+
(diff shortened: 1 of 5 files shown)


packages/card/person-card/index.tsx
```diff
@@ -18,13 +18,13 @@ interface PersonCardProps {
...
-    icon = <IconTwitter16 />
+    icon = <IconTwitter16 data-testid={'wpl-personcard-twitter-icon'} />
   }
   if (url.includes('github')) {
-    icon = <IconGithub16 />
+    icon = <IconGithub16 data-testid={'wpl-personcard-github-icon'} />
   }
   if (url.includes('linkedin')) {
-    icon = <IconLinkedin16 />
+    icon = <IconLinkedin16 data-testid={'wpl-personcard-linkedin-icon'} />
```
EnMod, line 27: 🤔 This feels...goofy to me? I was looking for a way to single out each of these icons in the test to ensure they each appear when the corresponding links are passed to the component. 

Perhaps I should rethink the test? 🤔

packages/card/event-card/index.test.js
```diff
@@ -0,0 +1,37 @@
...
+describe('<EventCard />', () => {
+  // Prop tests
+  it('should render the provided title prop', () => {
+    render(<EventCard {...defaultProps} />)
+
+    const title = screen.getByText(defaultProps.title)
+    expect(title.tagName).toBe('H2')
+  })
+
+  it('should render the provided eventType and eventDate prop together', () => {
```
EnMod, line 24: 🤔 I was considering somehow testing for the `eventDate` and `eventType` appearing in that order, but that seemed like it would be unnecessary. Thoughts?

alexcarpenter: yeah, not to worried about testing order here.

alexcarpenter (approved):

packages/card/person-card/index.tsx
EnMod: Will address this in a later PR if necessary!

PR #783, opened 2022-10-27
[Assembly] `<VideoFeature />`

EnMod:
🎟️ [Asana Task](https://app.asana.com/0/1201874072258661/1202983318984725/f)
🔍 [Preview Link](https://react-components-git-asm-videofeature-hashicorp.vercel.app/components/video-feature)

---

## Description

This is an assembly collecting changes for a new component: `<VideoFeature />` 

### PR Checklist 🚀

Items in this checklist may not may not apply to your PR, but please consider each item carefully.

- [x] Add Asana and Preview links above.
- [x] Conduct thorough self-review.
- [x] Add or update tests as appropriate.
- [x] Conduct reasonable cross browser testing for both compatibility and responsive behavior (We have a [Sauce Labs](https://app.saucelabs.com/) account for this, if you don't have access, just ask!).
- [ ] Conduct reasonable accessibility review (use the [WAS](https://accessible.org/Web-Accessibility-Standards-WAS-2.pdf) as a guide or an [axe browser plugin](https://www.deque.com/axe/) until we establish more formal checks).
- [x] Identify (in the description above) and document (add Asana tasks on [this board](https://app.asana.com/0/1100423001970639/list)) any technical debt that you're aware of, but are not addressing as part of this PR.

PR #784, opened 2022-10-27
feat(component): `<VideoFeature />`

EnMod:
🎟️ [Asana Task](https://app.asana.com/0/0/1203253460638593/f)
🔍 [Preview Link](https://react-components-git-nqvf-build-component-hashicorp.vercel.app/component/videofeature)

---

## Description

In this PR, the new `VideoFeature` component will be built out:
- [x] Props document updated
- [x] `docs.mdx` updated to reflect component props and showcase the component
- [x] Tests developed just enough to pass; further testing work will be handled in a future PR merged to assembly

### PR Checklist 🚀

### DIFF
--- a/packages/video-feature/index.tsx
+++ b/packages/video-feature/index.tsx
@@ -29,7 +29,7 @@ export default function VideoFeature({
       >
         <p>{'<InlineVideo />'}</p>
         <p>{video.url}</p>
-        <p>{video.caption}</p>
+        <p>{video.description}</p>
       </div>
     </div>
   )
--- a/packages/video-feature/index.tsx
+++ b/packages/video-feature/index.tsx
@@ -25,7 +25,7 @@ export default function VideoFeature({
       </div>
 
       <div className={s.video}>
-        <InlineVideo {...video} />
+        <InlineVideo {...video} gradientSide={contentSide} />
       </div>
     </div>
   )
--- a/packages/video-feature/index.tsx
+++ b/packages/video-feature/index.tsx
@@ -27,7 +27,11 @@ export default function VideoFeature({
       </div>
 
       <div className={s.video}>
-        <InlineVideo {...video} gradientPosition={contentPosition} />
+        <InlineVideo
+          {...video}
+          gradientPosition={contentPosition}
+          appearance={appearance}
+        />
       </div>
     </div>
   )
--- a/packages/video-feature/index.tsx
+++ b/packages/video-feature/index.tsx
@@ -1,12 +1,36 @@
 import classNames from 'classnames'
+import AuthorByline from '@hashicorp/react-author-byline'
+import Intro from '@hashicorp/react-intro'
 import { VideoFeatureProps } from './types'
 import s from './style.module.css'
 
 export default function VideoFeature({
-  children,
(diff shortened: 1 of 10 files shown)


packages/video-feature/types.ts
```diff
@@ -0,0 +1,12 @@
...
+import type { IntroProps } from '@hashicorp/react-intro/types'
+import type { InlineVideoProps } from '@hashicorp/react-inline-video/types'
+
+export interface VideoFeatureProps {
+  appearance?: 'light' | 'dark'
+  contentPosition?: 'left' | 'right'
+  heading: IntroProps['heading']
+  description: IntroProps['description']
+  author?: AuthorBylineProps
+  video: InlineVideoProps
```
alexcarpenter, line 11: Should we be picking the props we need here vs passing all of them along?

Wouldn'nt assume we expose appearance for author or video as those come from video.

Video also does not need to expose the gradientPosition as that is inferred from the contentPosition etc.

packages/video-feature/style.module.css
```diff
@@ -0,0 +1,43 @@
+.videoFeature {
+  --padding-y: 48px;
+  --padding-x: 48px;
+  --background-color: var(--wpl-neutral-100);
+
+  padding: var(--padding-y) var(--padding-x);
+  background-color: var(--background-color);
+  display: flex;
+  flex-direction: column-reverse;
```
alexcarpenter, line 9: flex usage is causing some layout issues here. Can we use grid?

packages/video-feature/types.ts
EnMod: Good points all around! I've used `Pick` instead to more specifically assign the type here:
https://github.com/hashicorp/react-components/pull/784/commits/98fe500d8415fff17d0f358ebb77b764e814ba0d

packages/video-feature/style.module.css
EnMod: Update: I've refactored the layout from `flex` to `grid` here. Some nice gains in readability, without the sizing/layout issues from the `flex` layout:
https://github.com/hashicorp/react-components/pull/784/commits/8cc180319ce97c319fada6d4e3184c641d068323

packages/video-feature/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+  author,
+  video,
+}: VideoFeatureProps) {
+  return (
+    <div
+      className={classNames(s.videoFeature, s[appearance], s[contentPosition])}
+    >
+      <div className={s.content}>
+        <Intro
+          heading={heading}
```
alexcarpenter, line 22:
```suggestion
          heading={heading}
          headingSize={3}
```

EnMod: Good catch!

EnMod:
@alexcarpenter I've tackled tests in a separate PR (#794). Plans going forward for this component:

1. Merge to the `<VideoFeature />` assembly when approved (#783 , it autoclosed after a rebase, please disregard)
2. Rebase and merge the tests PR (#794) into assembly when approved
3. Add the changeset in assembly

alexcarpenter (approved):

PR #789, opened 2022-10-28
feat(component): `<InlineVideo />`

EnMod:
🎟️ [Asana Task](https://app.asana.com/0/0/1203260018123853/f)
🔍 [Preview Link](https://react-components-git-nq-inlinevideo-hashicorp.vercel.app/components/inlinevideo)

---

## Description

This PR creates an `<InlineVideo />` component, and bumps `mktg-global-styles` to `4.3.1` so the component can properly render background gradients. Will be used soon in #784 .

### Tech debt

Testing for this PR and #784 will be handled in a separate PR after both are merged to the assembly for these components. 

### PR Checklist 🚀

### DIFF
--- a/packages/inline-video/index.test.js
+++ b/packages/inline-video/index.test.js
@@ -1,0 +1,16 @@
+import { render, screen } from '@testing-library/react'
+import InlineVideo from '.'
+
+const defaultProps = {
+  url: 'https://youtu.be/nJ0aI6sEDRo',
+  description:
+    'Description - 2 lines with character limit of 90. Tempus in egestas sagittis nulla feugiat',
+}
+
+describe('<InlineVideo />', () => {
+  it('TODO: Write tests', () => {
+    render(<InlineVideo {...defaultProps} />)
+    const element = screen.getByText(defaultProps.description)
+    expect(element).toBeInTheDocument()
+  })
+})
--- a/packages/inline-video/index.test.js
+++ b/packages/inline-video/index.test.js
@@ -8,7 +8,7 @@ const defaultProps = {
 }
 
 describe('<InlineVideo />', () => {
-  it('TODO: Write tests', () => {
+  it('renders the provided description text', () => {
     render(<InlineVideo {...defaultProps} />)
     const element = screen.getByText(defaultProps.description)
     expect(element).toBeInTheDocument()
--- a/packages/inline-video/index.tsx
+++ b/packages/inline-video/index.tsx
@@ -1,0 +1,25 @@
+import classNames from 'classnames'
+import type { InlineVideoProps } from './types'
+import s from './style.module.css'
+
+export default function InlineVideo(props: InlineVideoProps) {
+  const {
+    appearance = 'light',
+    gradientSide = 'left',
+    url,
+    description,
+    solution,
+  } = props
+  return (
+    <div
+      className={classNames(s.inlineVideo, s[appearance], s[gradientSide])}
+      data-testid="wpl-inline-video"
+    >
+      {url}
+      <br />
+      {description}
+      <br />
+      {solution}
+    </div>
+  )
+}
--- a/packages/inline-video/index.tsx
+++ b/packages/inline-video/index.tsx
@@ -19,7 +19,7 @@ export default function InlineVideo(props: InlineVideoProps) {
 
       <div className={s.videoPlayer}>{url}</div>
 
-      <p className={s.description}>{description}</p>
+      {description ? <p className={s.description}>{description}</p> : null}
     </div>
   )
 }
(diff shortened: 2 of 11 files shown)


packages/inline-video/style.module.css
```diff
@@ -0,0 +1,79 @@
...
+    --gradient-stops: #f26c64 0%, #ffe68d 100%;
+  }
+
+  &.networking {
+    --gradient-stops: #c750b7 0%, #ffb3b7 100%;
+  }
+
+  &.applications {
+    --gradient-stops: #5493e4 0%, #5bd0d9 50%, #a3f7d7 100%;
+  }
```
alexcarpenter, line 54: these are available via wpl tokens https://github.com/hashicorp/mktg-global-styles/blob/main/custom-properties/color.css#L132-L191

packages/inline-video/style.module.css
```diff
@@ -0,0 +1,79 @@
...
+  @media (--medium-up) {
+    @nest .left & {
+      padding-left: var(--gradient-size);
+    }
+
+    @nest .right & {
+      padding-right: var(--gradient-size);
+    }
+
+    &::before {
```
alexcarpenter, line 20: Thoughts on just making this gradient the full width of the element and changing the padding left/right based on placement?

EnMod: Ah, I think I know what you mean! I'll give it a go and see how that looks

EnMod: It ended up working well! The changes are captured in the latest commit (along with a fix for a regression 😅)

packages/inline-video/style.module.css
EnMod: @alexcarpenter Hmm...I've replaced all the values here with their proper tokens, and yet the actual declaration doesn't apply the value initially assigned to `--gradient`? Changing the `solution` prop to a supported value _does_ allow the correct gradient to show, however. 🤔 

Maybe I'm running into some kind of scoping or syntax issue?

EnMod: Update: Just needed to update deps! We're good to go as of the latest commit:
https://github.com/hashicorp/react-components/pull/789/commits/65a6e3c9295c7a3fed292c039321bc5ae3469648

packages/inline-video/types.ts
```diff
@@ -0,0 +1,7 @@
+export interface InlineVideoProps {
+  appearance?: 'light' | 'dark'
+  url: string
+  description?: string
+  solution?: 'infrastructure' | 'security' | 'networking' | 'applications'
+  gradientSide?: 'left' | 'right'
```
alexcarpenter, line 6: 💭 How about `gradientPosition` vs `gradientSide`?

packages/inline-video/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  }
+  return (
+    <div
+      className={classNames(s.inlineVideo, s[appearance], s[gradientSide])}
+      data-testid="wpl-inline-video"
+    >
+      <div className={classNames(s.videoContainer, solution && s[solution])}>
+        <div className={s.video}>
+          {url.includes('wistia') ? (
+            <WistiaPlayer {...playerProps} />
```
alexcarpenter, line 30: On wistia videos, I think we want to hide controls on load until video is played. Ref: https://github.com/hashicorp/hashicorp-www-next/blob/main/pages/about/index.tsx#L198-L202

```js
config={{
  options: {
    controlsVisibleOnLoad: false,
  },
}}
```

alexcarpenter: CleanShot 2022-11-01 at 13 16 34@2x

EnMod: Good call, makes sense to keep that behavior consistent :+1: 

I've addressed this here:
https://github.com/hashicorp/react-components/pull/789/commits/4decdef0a19f180150dfe468ff095216e2c9e5e3

packages/inline-video/types.ts
EnMod: I like it! Addressed here:
https://github.com/hashicorp/react-components/pull/789/commits/3ad996e88610cc6d7b68231f208620674a254ba1

packages/inline-video/index.tsx
EnMod: Good point! I've addressed that here:
https://github.com/hashicorp/react-components/pull/789/commits/4decdef0a19f180150dfe468ff095216e2c9e5e3

packages/inline-video/index.test.js
```diff
@@ -0,0 +1,16 @@
...
+import InlineVideo from '.'
+
+const defaultProps = {
+  url: 'https://youtu.be/nJ0aI6sEDRo',
+  description:
+    'Description - 2 lines with character limit of 90. Tempus in egestas sagittis nulla feugiat',
+}
+
+describe('<InlineVideo />', () => {
+  it('TODO: Write tests', () => {
```
alexcarpenter, line 11:
```suggestion
  it('renders the provided description text', () => {
```

alexcarpenter (approved):
LGTM, I think we can just update the test name to reflect checking the description presence and we're good for an initial release.

PR #793, opened 2022-11-02
feat(consent manager): Add function to check if preferences are loaded

pbortnick:
🎟️ [Asana Task](https://app.asana.com/0/1202097197789424/1203274026595792/f)

---

## Description

Exports a const that lets us know if preferences have been loaded by reading a global variable set within `loadPreferences()`. The motivation here is to use this in order to prevent re-runs of `saveAndLoadAnalytics()` in our web properties. See [this draft PR for example](https://github.com/hashicorp/dev-portal/pull/1317/files#diff-b359798aa474d6e3d6c2a75fe11ed4d0bb12fed48592453f433524ed0be5ab5dR52). In this example, we want to run `saveAndLoadAnalytics()` only after sign in. So here we'd replace `if (!loadPreferences())` to `if (!preferencesSavedAndLoaded())`. It is essentially the same check but it provides the data without having to re-read the cookie, optimizing performance.

### PR Checklist 🚀

### DIFF
--- a/packages/consent-manager/util/cookies.js
+++ b/packages/consent-manager/util/cookies.js
@@ -50,3 +50,12 @@ export function savePreferences(prefs, version) {
     domain,
   })
 }
+
+export function preferencesSavedAndLoaded() {
+  const preferences = loadPreferences()
+  if (preferences && preferences.length > 0) {
+    return true
+  }
+
+  return false
+}
--- a/packages/consent-manager/util/cookies.js
+++ b/packages/consent-manager/util/cookies.js
@@ -43,12 +43,13 @@ export function getDomain() {
 
 export function loadPreferences() {
   const cookiesJSON = cookies.getJSON(COOKIE_KEY)
-  if (
-    cookiesJSON &&
-    Object.keys(
-      typeof cookiesJSON === 'string' ? JSON.parse(cookiesJSON) : cookiesJSON
-    ).length > 0
-  ) {
+  if (cookiesJSON) {
+    // if (
+    //   cookiesJSON &&
+    //   Object.keys(
+    //     typeof cookiesJSON === 'string' ? JSON.parse(cookiesJSON) : cookiesJSON
+    //   ).length > 0
+    // ) {
     preferencesLoaded = true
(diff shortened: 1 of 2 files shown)


zchsh (approved):

packages/consent-manager/util/cookies.js
```diff
@@ -40,7 +42,14 @@ export function getDomain() {
 }
 
 export function loadPreferences() {
-  return cookies.getJSON(COOKIE_KEY)
+  const cookiesJSON = cookies.getJSON(COOKIE_KEY)
+  if (cookiesJSON) {
+    preferencesLoaded = true
+  } else if (preferencesLoaded) {
+    preferencesLoaded = false
+  }
```
ashleemboyer, line 50: Question 🙋🏻 I'm not picturing when we would need to set `preferencesLoaded` to `false` if `preferencesLoaded` is already `true`. Is there a specific case where we need that to happen?

packages/consent-manager/util/cookies.js
```diff
@@ -49,4 +58,7 @@ export function savePreferences(prefs, version) {
     expires: COOKIE_EXPIRES,
     domain,
   })
+  preferencesLoaded = true
 }
+
+export const preferencesSavedAndLoaded = () => !!preferencesLoaded
```
ashleemboyer, line 64: Optional 🎨 If we default `preferencesLoaded` to `false` when it's initialized above `getDomain`, I think we'd just be able to return the variable from this function without the `!!`.

packages/consent-manager/util/cookies.js
pbortnick: @ashleemboyer good point. I added as a safeguard but yeah no scenarios that I can think of so it is probably unnecessary.

packages/consent-manager/util/cookies.js
pbortnick: Good point! I had it as undefined originally when writing some tests but we can go this route now 👍

ashleemboyer (approved):

PR #802, opened 2022-11-07
chore(deps): upgrade to next v12.3

dstaley:
🎟️ [Asana Task]()

---

## Description

This PR upgrades the repo from Next v11 😅 to Next v12.3, an important stepping-stone to eventually upgrading to Next v13.

Changes:

- Small tweak to Jest mocks for `useRouter`
- Don't test `async: false` case for consent manager scripts (`async: false` causes scripts to be emitted during the server rendering phase, and are thus untestable in Jest since it only uses client-side rendering)
- Tweak URL param logic for `marketo-form` to handle the new type (`string | string[] | undefined`)

### PR Checklist 🚀

### DIFF
--- a/.changeset/ninety-eyes-shout.md
+++ b/.changeset/ninety-eyes-shout.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-marketo-form': patch
+---
+
+Throw error if form query parameter is missing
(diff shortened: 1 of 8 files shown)


BRKalow (approved):
Looks great, thank you! 😄

PR #810, opened 2022-11-08
refactor: card primitive

alexcarpenter:
🔍 [Preview Link](https://react-components-git-accard-refactor-hashicorp.vercel.app/components/card)

---

## Description

Refactors Card to remove the optional children or props approach and exports card primitives to be used to build up our different card variant needs.

The base card variant should be unaffected in use on dotcom and within RelatedContent.

### PR Checklist 🚀


packages/card/docs.mdx
```diff
@@ -215,72 +122,34 @@ A component used to promote and link to marketing pages, always using a descript
...
-
-<PropsTable props={componentProps.metaProps} />
-
-### `<Heading />`
-
-<PropsTable props={componentProps.headingProps} />
-
-### `<Description />`
-
-<PropsTable props={componentProps.descriptionProps} />
```
EnMod, line 286: Should these prop docs be removed?

alexcarpenter: They were, I think your commenting on code that was deleted.

EnMod: Right, I'm asking if they _should_ be removed :+1: I was thinking we should have prop docs in this file for the component?

alexcarpenter: ah I see what you mean, I can bring them back if you'd like. I though folks would not be interacting with these sub components outside of building the variants within react-components.

alexcarpenter: ah yeah, good call! https://github.com/hashicorp/react-components/pull/810/commits/4deaaaf47bc92c5b6f282ea1be4877f14a2c36cf

EnMod (approved):
🚀 🚀 🚀

PR #828, opened 2022-11-23
Enable default autofill values for hidden fields

nandereck:
🎟️ [Asana Task]()

---

## Description
Currently, values for hidden fields can only be derived from query parameters. Both the Newsletter Signup Form and Certification mailing list form have hidden fields with default values defined in Marketo and these values need to be passed through when the form is submitted.

This PR adds a check within `<HiddenField />` that sets the value for the field if a default value is provided.

### PR Checklist 🚀

### DIFF
--- a/.changeset/giant-eels-taste.md
+++ b/.changeset/giant-eels-taste.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-marketo-form': minor
+---
+
+Allow for hidden autofill values to be passed through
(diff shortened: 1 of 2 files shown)


dstaley:
Just to clarify: this adds support for defining the default value for a hidden field via Marketo, correct?

nandereck:
> Just to clarify: this adds support for defining the default value for a hidden field via Marketo, correct?

Yes. This will allow one to create a hidden field in Marketo with a default value.

dstaley (approved):

PR #834, opened 2022-11-30
fix(code-block): only render HiddenCopyContent after hydration

BRKalow:
🎟️ [Asana Task]()

---

## Description

Adjust `HiddenCopyContent` to only render after hydration as occurred. This helps avoid some tricky hydration issues, and also avoids duplicative markup being shipped from the server. As the element is only used by the copy-to-clipboard interaction, it's safe to defer rendering.

### PR Checklist 🚀

### DIFF
--- a/packages/code-block/partials/hidden-copy-content/index.js
+++ b/packages/code-block/partials/hidden-copy-content/index.js
@@ -1,4 +1,4 @@
-import { forwardRef } from 'react'
+import { forwardRef, useEffect, useState } from 'react'
 
 /**
  * This hidden element acts solely as a container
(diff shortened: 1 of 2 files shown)


thiskevinwang (approved):
Side: i wonder if a helper-wrapper — something like Next's `dynamic( ..., { ssr: false})` — would be a useful utility for this Client-only guard

packages/code-block/partials/hidden-copy-content/index.js
```diff
@@ -8,6 +8,19 @@ import { forwardRef } from 'react'
  * line numbers and so on in other parts of code-block.
  */
 function HiddenCopyContent({ code }, copyRef) {
+  const [isClient, setIsClient] = useState(false)
+
+  useEffect(() => {
+    setIsClient(true)
+  }, [])
```
thiskevinwang, line 15: General Q: would this accomplish the same end result?

```suggestion
  if (typeof window === "undefined") return null
```

BRKalow: Nope, because the initial client render would then be different from the server render and we could still encounter hydration issues.

BRKalow: https://beta.reactjs.org/apis/react-dom/hydrate#handling-different-client-and-server-content

PR #837, opened 2022-11-30
Render `react-player` only if the `window` exists

nandereck:
🎟️ [Asana Task]()

---

## Description

Unless `react-player` is loaded on the client-side using dynamic/no-ssr or `useEffect`, React will panic with a hydration error (see [this issue](https://github.com/cookpete/react-player/issues/1474)). 

This PR updates the `InlineVideo` component to conditionally render the player if the `window` not `undefined`.

### PR Checklist 🚀

### DIFF
--- a/.changeset/red-jokes-allow.md
+++ b/.changeset/red-jokes-allow.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-inline-video': minor
+---
+
+Add lazy loading to prevent hydration errors
--- a/packages/inline-video/index.tsx
+++ b/packages/inline-video/index.tsx
@@ -1,8 +1,8 @@
 import classNames from 'classnames'
 import WistiaPlayer from 'react-player/wistia'
+import Player from 'react-player'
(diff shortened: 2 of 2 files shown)


alexcarpenter (approved):

BRKalow:
Just a heads up that this will cause layout shift, depending on where the component is used. For the video player on the developer homepage, we created a placeholder component that renders before the video player mounts to reserve space.

https://github.com/hashicorp/dev-portal/blob/main/src/components/video-embed/index.tsx#L28

nandereck:
> Just a heads up that this will cause layout shift

Is that the case since we explicitly set the [aspect ratio ](https://github.com/hashicorp/react-components/blob/58adf3d7befb799c7351f9341f1bcf4e84b36b26/packages/inline-video/style.module.css#L50) for the container?

BRKalow:
@nandereck Ooh awesome, I missed that we _were_ in fact rendering a shell. My bad!

PR #848, opened 2022-12-13
feat(react-card): add custom cards for faceted search

pbortnick:
🎟️ [Asana Task](https://app.asana.com/0/1203227460033471/1203451232305046/f)
🔍 [Preview Link](https://react-components-git-pbadd-custom-cards-hashicorp.vercel.app/components/card)

---

## Description

Add more custom cards that are used in faceted search pages
- blog
- career
- customer story
- newsroom
- partner
- resource

### PR Checklist 🚀

### DIFF
--- a/packages/card/career-card/style.module.css
+++ b/packages/card/career-card/style.module.css
@@ -7,5 +7,5 @@
 .subHeading {
   composes: g-type-body-small-x-strong from global;
   margin: 0;
-  color: var(--wpl-neutral-500);
+  color: var(--secondary-text-color);
 }
--- a/packages/card/career-card/style.module.css
+++ b/packages/card/career-card/style.module.css
@@ -1,0 +1,11 @@
+.bottom {
+  display: flex;
+  align-items: flex-end;
+  height: 100%;
+}
+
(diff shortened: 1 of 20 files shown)


alexcarpenter (commented):
Looking good, I don't think we need the bottom class usage you've added. To my knowledge the badges are not bottom aligned which you can see from current implementations. They just happen to be aligned I think because the text is all the same length in Figma examples.

CleanShot 2022-12-14 at 13 57 44@2x

packages/card/career-card/style.module.css
```diff
@@ -0,0 +1,11 @@
+.bottom {
+  display: flex;
+  align-items: flex-end;
+  height: 100%;
+}
+
+.subHeading {
+  composes: g-type-body-small-x-strong from global;
+  margin: 0;
+  color: var(--wpl-neutral-500);
```
alexcarpenter, line 10: looks like we can use `var(--secondary-text-color)` here vs neutral-500

alexcarpenter (approved):
LGTM, just one color update we should make.

PR #862, opened 2023-01-11
Update `<Badge />`

nandereck:
🎟️ [Asana Task](https://app.asana.com/0/1202738955551191/1203644233223829/f)
🔍 [Preview Link](https://react-components-git-naupdatebadge-hashicorp.vercel.app/components/badge)

---

## Description
This PR updates `<Badge />` to match the updated [WPL spec](https://www.figma.com/file/7cYgDM618stjYUHDqAfRec/Components?node-id=1361%3A6862&t=ZTuBbEbE9ORI5bON-1). The biggest call-out here is around the complexity of the implementation. The spec calls for 120 variations of the `Badge`. Due to the variability between variations, it's really difficult to implement this in a programmatic way. :(

## Validation Steps
- Navigate to the example section in the deploy preview
- Validate that the table of badges matches the [WPL spec](https://www.figma.com/file/7cYgDM618stjYUHDqAfRec/Components?node-id=1361%3A6862&t=ZTuBbEbE9ORI5bON-1).

### PR Checklist 🚀


packages/badge/docs.mdx
```diff
@@ -6,36 +6,74 @@ Badges are non-interactive labels highlighting metadata such as status or catego
...
-    <Badge theme="white" variant="secondary">New</Badge>
-  </div>
-</>
 `}
 </LiveComponent>
 
 <PropsTable props={componentProps} />
+
+## Examples
+
```
nandereck, line 18: This is pretty wild but it does render every possible variation of the `<Badge />`. Not sure how else we would be able to QA.

packages/badge/index.tsx
```diff
@@ -1,23 +1,47 @@
...
+    'terraform',
+    'vault',
+    'vagrant',
+    'waypoint',
+  ]
+  const isProductTheme = products.includes(theme)
+  const productName = isProductTheme ? theme : undefined
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore - The ternary above ensures that `useProductMeta` always gets
+  // the appropriate value but this does not seem to appease the the TS gods
```
nandereck, line 26: A bit stumped by this TS error. Would love feedback so as to avoid `ts-ignore`'ing.

alexcarpenter (commented):
Looking good!

Couple of suggestions to alleviate the need to use ts-ignore.

```ts
// badge/types.ts
import type { Products } from '@hashicorp/platform-product-meta'

// Extract types to use within the badge component
export type ProductThemes = Exclude<Products, 'hashicorp'>
export type ExtraThemes = 'neutral' | 'action'
export type BadgeThemes = ProductThemes | ExtraThemes

export interface BadgeProps {
  children: string
  variant?: 'primary' | 'secondary'
  theme?: BadgeThemes
  page?: 'light' | 'faint' | 'strong' | 'strongFaint' | 'action' | 'actionFaint'
}

```

```ts
// badge/index.tsx
import type { BadgeProps, ProductThemes } from './types'
import useProductMeta from '@hashicorp/platform-product-meta'

const Badge = ({
  children,
  variant = 'primary',
  theme = 'vagrant',
  page = 'light',
}: BadgeProps) => {
  let productName: ProductThemes | undefined
  // typescript doesn't work great with includes
  if (theme === 'neutral' || theme === 'action') {
    productName = undefined
  } else {
    productName = theme
  }
  const { themeClass: productThemeClass } = useProductMeta(productName)
  // productThemeClass will return undefined if it is passed undefined which happens when the theme is neutral or action
  const themeClass = productThemeClass
    ? `${productThemeClass} ${s.theme_product}`
    : s[`theme_${theme}`]

  return (
    <span
      className={classNames([
        s.root,
        s[variant],
        themeClass,
        s[`page_${page}`],
      ])}
    >
      {children}
    </span>
  )
}

export default Badge

```

nandereck:
> Couple of suggestions to alleviate the need to use ts-ignore.

Amazing, thanks! This is very clarifying. 👍

.changeset/twenty-rocks-end.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hashicorp/react-badge': major
+---
```
nandereck, line 3: @alexcarpenter - I was thinking about updating the changeset to a `minor` update. Looking at the old component, it doesn't introduce any breaking changes. Does that sound reasonable?

alexcarpenter: Works for me. Should be an easy update if needed where it is being used currently.

packages/badge/index.tsx
```diff
@@ -1,23 +1,37 @@
...
+const Badge = ({
   children,
   variant = 'primary',
-  theme = 'neutral',
-}: BadgeProps) {
-  let themeClass
-  const productClass = productStyles[theme]
-  if (productClass) {
-    themeClass = classNames(s.product, productClass)
+  theme = 'vagrant',
```
alexcarpenter, line 9: Can we default to action here?

nandereck: Done! 👍

PR #881, opened 2023-02-15
fix(EnterpriseAlert): render span when inline

BRKalow:
🎟️ [Asana Task]()

---

## Description

When used inline, `EnterpriseAlert` currently renders a `div` as its root element. This is problematic when used in MDX specifically, as it will most likely be a child of a `p`. Per the spec, `p` elements cannot contain block-level elements and so the HTML is invalid. This results in React hydration errors.

I've updated `EnterpriseAlert` to render `span` as its root element if the `inline` prop is passed to fix this.

### PR Checklist 🚀

### DIFF
--- a/packages/enterprise-alert/index.tsx
+++ b/packages/enterprise-alert/index.tsx
@@ -23,8 +23,12 @@ function EnterpriseAlert({
   className,
 }: EnterpriseAlertProps) {
   const { name, slug, themeClass } = useProductMeta(product)
+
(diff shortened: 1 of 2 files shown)


packages/enterprise-alert/index.tsx
```diff
@@ -23,8 +23,12 @@ function EnterpriseAlert({
   className,
 }: EnterpriseAlertProps) {
   const { name, slug, themeClass } = useProductMeta(product)
+
+  // This ensures we aren't producing invalid HTML when rendering inline alerts within MDX. When used inline, we might end up nesting a div inside of a p. This is invalid as p cannot contain block-level elements (ref: https://www.w3.org/TR/html401/struct/text.html#h-9.3.1).
+  const Element = inline ? 'span' : 'div'
```
thiskevinwang, line 28: TIL 💯

thiskevinwang (approved):

PR #893, opened 2023-03-01
fix(alert-banner): ensure alert banner doesn't flash if previously dismissed

BRKalow:
🎟️ [Asana Task]()

---

## Description

### PR Checklist 🚀

### DIFF
--- a/packages/alert-banner/index.tsx
+++ b/packages/alert-banner/index.tsx
@@ -1,8 +1,3 @@
-/**
- * Copyright (c) HashiCorp, Inc.
- * SPDX-License-Identifier: MPL-2.0
- */
-
 import { useEffect, useState } from 'react'
 import cookie from 'js-cookie'
 import slugify from 'slugify'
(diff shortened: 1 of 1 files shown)


packages/alert-banner/index.tsx
```diff
@@ -71,32 +80,54 @@ function AlertBanner({
...
-        <VisuallyHidden>Dismiss alert</VisuallyHidden>
-      </button>
-    </div>
+        </a>
+        <button className={s.closeButton} onClick={closeBanner}>
+          <InlineSvg src={CloseIcon} />
+          <VisuallyHidden>Dismiss alert</VisuallyHidden>
+        </button>
+      </div>
+      <script
```
dstaley, line 113: @BRKalow In what instance is this inline script run? Is this to handle when the server-side HTML contains the alert banner but it's been dismissed by the user? If that's the case, shouldn't we switch the logic such that the server always returns the alert hidden, and it then is shown once the `useState` hook value is calculated on the client?

kendallstrautman: @dstaley would that cause another layout shift though when we try to render the alert?

kendallstrautman (approved):
Tested on https://github.com/hashicorp/dev-portal/pull/1764 and this seems to resolve the issue.

packages/alert-banner/index.tsx
BRKalow: @dstaley yes that scenario is correct. wouldn't we have layout shift still? With this check, the isShown class is removed before the browser paints and so it removes the shift. If we wait until the state value is initialized there would be a shift I think.

dstaley: Ahhh, okay I see what's happening here! I'm not thrilled at the idea of a render-blocking inline script, but I agree that's preferred to layout shift.

dstaley (approved):

packages/alert-banner/index.tsx
BRKalow: @dstaley Yeah in general I agree, but in this scenario it seems like a reasonable tradeoff 🤞

PR #897, opened 2023-03-07
chore: updates next/image uses to next/legacy/image

BRKalow:
🎟️ [Asana Task]()

---

## Description

Update react-components `next/image` uses to `next/legacy/image` for full compatibility with next 13 now that all of our sites have been upgraded.

### PR Checklist 🚀

---
  - https://app.asana.com/0/0/1203354752975570


.changeset/nasty-pans-hug.md
```diff
@@ -0,0 +1,11 @@
+---
+'@hashicorp/react-code-block': major
+'@hashicorp/react-docs-page': major
+'@hashicorp/react-hero': major
+'@hashicorp/react-author-byline': minor
+'@hashicorp/react-card': minor
+'@hashicorp/react-marketo-form': minor
+'@hashicorp/react-notification': minor
```
BRKalow, line 8: The `minor` updates here are pre-1.0 packages, so this should be synonymous with a major per https://semver.org/#spec-item-4

> Version 1.0.0 defines the public API. The way in which the version number is incremented after this release is dependent on this public API and how it changes.

next-env.d.ts
```diff
@@ -1,8 +1,3 @@
-/**
- * Copyright (c) HashiCorp, Inc.
- * SPDX-License-Identifier: MPL-2.0
- */
-
```
dstaley, line 5: is the bot gonna get angry with us for removing this?

packages/card/primitives.tsx
```diff
@@ -5,7 +5,7 @@
 
 import * as React from 'react'
 import classNames from 'classnames'
-import Image from 'next/image'
```
dstaley, line 8: I think `<Card>` actually needs `next/image` since @alexcarpenter fixed the CSS.

dstaley (approved):
LGTM besides the `<Card>` which I believe works on `next/image`.

next-env.d.ts
BRKalow: Probably 👀 I think we can define some config to exclude this file, I'll do that in a follow up!

BRKalow: https://github.com/hashicorp/react-components/pull/898

packages/card/primitives.tsx
BRKalow: Ah, great catch! I will revert this then 👍

PR #905, opened 2023-03-16
fix(placeholder): Shine height on Safari/iOS

EnMod:
🔍 [Preview Link](https://react-components-git-nqplaceholder-fix-shine-height-hashicorp.vercel.app/components/placeholder)

---

## Description

This PR fixes an issue with `.box` `background-size` causing the "shine" animation not to fill the box height on Safari/iOS.

Context: `.box` had no explicit height value set for its `background-size` property, and values default to auto if not set. However, in Safari/iOS, the `auto` for the height value is not the height of the container, but (from observation) the value set for width. This behavior seems to have caused the "shine" to only reach `75%` of the box's height in Safari/iOS. Setting the height value to an explicit `100%` mitigates the issue.

### PR Checklist 🚀

### DIFF
--- a/.changeset/stale-plants-divide.md
+++ b/.changeset/stale-plants-divide.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/react-placeholder': patch
+---
+
+Fixes shine animation height not filling its containing box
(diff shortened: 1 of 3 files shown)


BRKalow (approved):
Makes sense to me, good catch!

PR #989, opened 2023-08-01
chore: remove @reach/auto-id dep, opt for react's `useId`

kendallstrautman:
🎟️ [Asana Task]()
🔍 [Preview Link](https://react-components-git-ksupdate-use-id-hashicorp.vercel.app)

---

## Description

`@reach/auto-id` is pinned <= react 17 and this dependency causes conflicts / warnings where react 18 is used. This refactors all uses of `@reach/auto-id` to use the [react version](https://react.dev/reference/react/useId). 

Screenshot 2023-08-01 at 9 04 25 AM

### PR Checklist 🚀

### DIFF
--- a/.changeset/old-days-rescue.md
+++ b/.changeset/old-days-rescue.md
@@ -1,7 +1,7 @@
 ---
-'@hashicorp/react-checkbox-input': patch
-'@hashicorp/react-text-input': patch
-'@hashicorp/react-textarea-input': patch
+'@hashicorp/react-checkbox-input': major
+'@hashicorp/react-text-input': major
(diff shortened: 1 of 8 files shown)


thiskevinwang (changes requested):
We'll need to bump `peerDependencies.react` in each of the `package.json`s to `>=18.x` since [`useId` is only available in 18+](https://react.dev/blog/2022/03/29/react-v18#useid)

dstaley:
I think Dev Portal is the only repo using these packages, so it should be safe to update!

dstaley (approved):

thiskevinwang (approved):

PR #991, opened 2023-08-03
feat(code-block): add wrapCode option

zchsh:
🎟️ [Asana Task][task]
🔍 [Preview Link][preview]

---

This PR updates the `code-block` to add support for a `wrapCode` `boolean` option.

When enabled, the `wrapCode` option causes long lines of code within a code block to wrap to new lines, rather than create horizontal overflow within the code block.

### 🎨 Screenshots

| Default, overflowing | `options.wrapCode={true}` |
| - | - |
| overflow | wrap |

### 💻 Demo

https://github.com/hashicorp/react-components/assets/4624598/782b2710-abaf-45be-ba6f-062178441f2d

### 🧪 Testing

- [ ] Visit [the preview][preview] of changes in Swingset
    - The `wrapCode` example at the start of the `Features & Guidelines` section should look and function as expected, across viewport sizes
- [ ] Take a gander at the [downstream PR preview](https://dev-portal-git-zsopenapi-code-wrap-hashicorp.vercel.app/hcp/api-docs/vault-secrets) that uses the latest canary release from this work to allow URL paths in OpenAPI docs to wrap
    - https://github.com/hashicorp/dev-portal/pull/2122

### PR Checklist 🚀

    - I think snapshot tests would be most appropriate to validate the contents of this PR. Maybe one day Swingset will support them, or maybe we could explore other tooling to make this possible! In the past I think we had a setup with Storybook & Percy running 💭 
- [x] Conduct reasonable cross browser testing for both compatibility and responsive behaviour (We have a [Sauce Labs](https://app.saucelabs.com/) account for this, if you don't have access, just ask!).
    - I don't have access to Sauce Labs, but I've tested this work in Arc, Chrome, Firefox, and Safari.

[task]: https://app.asana.com/0/1204678746647847/1205233741731093/f
[preview]: https://react-components-git-zscode-block-support-wrap-hashicorp.vercel.app/


packages/code-block/docs.mdx
```diff
@@ -42,7 +45,54 @@ A line that goes on for a very long time so that it overflows the container in w
...
+
+Note this also works with line numbers and line highlighting.
+
+<CodeBlock
+  options={{ showClipboard: true, lineNumbers: true, highlight: '1,3' }}
+  theme="dark"
+  code={`A line that goes on for a very long time so that it overflows the container in which it is located, which might be a pretty wide container.\nThis is a second line of code.\nAnd a third line.\nAnd another line, this is the fourth line.`}
+/>
+
+#### Wrap Code
```
zchsh, line 59: Added a section to `docs.mdx` to demo the new behaviour.

packages/code-block/partials/code-lines/code-lines.module.css
zchsh, line 1: This file is likely easier to view in "side by side" mode, as the corresponding JSX has been significantly refactored.

packages/code-block/partials/code-lines/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+  let numberSpacer: string | null = null
+  if (lineNumbers) {
+    const padLevel = Math.max(linesOfCode.length.toString().length, 1)
+    numberSpacer = ''.padEnd(padLevel)
+  }
+
+  // When the floating copy button is present, we add padding to many lines
+  const padRight = Boolean(hasFloatingCopyButton)
+
+  if (wrapCode) {
```
zchsh, line 79: I've refactored this component fairly significantly, which felt necessary in order to make the `wrapCode` option feasible.

For some context, there are fundamentally different layout needs with `wrapCode` compared to without:
- When allowing overflow, we want the lines of code to scroll together, but we want the line numbers for each line to stay in a fixed position, like a frozen spreadsheet column. This necessitates a two-column layout at the top level, with the line numbers in one column, and the line contents in a separate adjacent column with scrollable content.
- When wrapping code, a single line will take up more vertical space when it wraps. The two-column layout needed for overflowing code is no longer usable - the only approach I can imagine would be to somehow detect wrapping in each line in the "lines" columns and constantly update the corresponding line number contents in the "numbers" column to contain an identical number of newlines. It seemed much easier to rethink this use case as an alternate layout of the same atomic components, with a single column where each "row" is a flex layout that contains both the line number and the line contents.

To achieve these varied layouts, the previous `CodeLines` component was first refactored into smaller composable components, and then here in this conditional, we compose those smaller components into the appropriate layouts depending on the incoming `wrapCode` boolean.

zchsh:
@dstaley Tagged ya here since while this PR and the `wrapCode` option it introduces is currently for a dev-dot use case, this'll obviously show up on a lot of other sites as well - would love another set of eyes on it 🙇‍♂️

packages/code-block/partials/code-lines/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+
+import React, { PropsWithChildren, ReactElement, useMemo } from 'react'
+import classNames from 'classnames'
+import parseHighlightedLines from '../../utils/parse-highlighted-lines'
+import splitJsxIntoLines from './utils/split-jsx-into-lines'
+import splitHtmlIntoLines from './utils/split-html-into-lines'
+import s from './code-lines.module.css'
+
+interface CodeLinesProps {
+  code: string | ReactElement
```
dstaley, line 23: Just curious: is there a specific reason you went with `ReactElement`? I _think_ `code: ReactNode` would be the more accurate type since that's inclusive of `string`.

packages/code-block/partials/code-lines/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+function CodeLines({
+  code,
+  language,
+  lineNumbers,
+  highlight,
+  hasFloatingCopyButton,
+  wrapCode,
+}: CodeLinesProps) {
+  // Parse out an array of integers representing which lines to highlight
+  const highlightedLines = parseHighlightedLines(highlight) as number[]
```
dstaley, line 44: Does this need to be cast?

dstaley (approved):
This is definitely a tricky problem, but I love that your solution is clear and thoroughly documented! Let two non-blocking comments.

packages/code-block/partials/code-lines/index.tsx
zchsh: Great point! Intent was that we might want `string | ReactElement` as it gets a little more specific about `code` not being any of the other possible types under `ReactNode`:

```ts
type ReactNode = ReactElement | string | number | ReactFragment | ReactPortal | boolean | null | undefined;
```

packages/code-block/partials/code-lines/index.tsx
zchsh: For now yes, as otherwise it seems the inferred type of `highlightedLines` ends up being `never[]`, and we get a type error later at `highlightedLines.indexOf(number) !== -1`.

Ideally I'd have jumped in and converted `parseHighlightedLines` to TypeScript as well, but I didn't quite have the energy 😅

PR #1001, opened 2023-09-06
feat(card): Make `date` optional on `<ResourceCard />`

EnMod:
🎟️ [Asana Task for request prompting the change](https://app.asana.com/0/1202791227659279/1205347737637312/f)
🔢 [PR on `www` implementing the canary release](https://github.com/hashicorp/web/pull/781)
🔍 [Preview Link](https://react-components-git-nqcard-make-dates-optiona-b3beed-hashicorp.vercel.app/components/card)
🔍 [Preview Link for example usage on `www` (see each case study library item's card)](https://hashicorp-www-next-git-nqwww-hide-case-study-l-583a57-hashicorp.vercel.app/case-studies)

---

## Description

This PR updates our `<ResourceCard />` component to make showing the date optional; if no date is passed in, no date (or `|` separator) is shown.

### PR Checklist 🚀

### DIFF
--- a/packages/card/resource-card/index.test.tsx
+++ b/packages/card/resource-card/index.test.tsx
@@ -18,7 +18,7 @@ const defaultProps = {
 }
 
 describe('<ResourceCard />', () => {
-  it('should render the provided metadata correctly', () => {
+  it('should render the provided date and category with separator', () => {
     const expectedMeta = [defaultProps.date, defaultProps.category]
 
     render(<ResourceCard {...defaultProps} />)
@@ -30,5 +30,19 @@ describe('<ResourceCard />', () => {
     })
 
     expect(metaElement).toContainElement(screen.getByText('|'))
+  })
+
+  it('should not render the date if no date is provided', () => {
+    defaultProps.date = ''
+
+    render(<ResourceCard {...defaultProps} />)
+
+    const metaElement = screen.getByTestId('wpl-card-meta')
+
+    expect(metaElement).toContainElement(
+      screen.getByText(defaultProps.category)
+    )
+
+    expect(metaElement).not.toContainElement(screen.queryByText('|'))
   })
 })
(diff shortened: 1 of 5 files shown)


packages/card/resource-card/index.test.tsx
```diff
@@ -31,4 +31,18 @@ describe('<ResourceCard />', () => {
 
     expect(metaElement).toContainElement(screen.getByText('|'))
   })
+
+  it('should not render the date if no date is provided', () => {
+    defaultProps.date = ''
```
dstaley, line 36: Instead of modifying default props, make a new object and spread defaultProps

```
const defaultPropsWithoutDate = { ...defaultProps, date: '' }
```

packages/card/resource-card/index.tsx
```diff
@@ -25,6 +25,12 @@ export function ResourceCard({
   thumbnail,
   appearance,
 }: ResourceCardProps): JSX.Element {
+  const meta: string[] = [category]
+
+  if (date) {
+    meta.unshift(date)
```
dstaley, line 31: Instead of mutating an array, let's use something like this:

```ts
const meta: string[] = date ? [date, category] : [category]
```

packages/card/resource-card/index.test.tsx
EnMod: Good call, addressed here:
https://github.com/hashicorp/react-components/pull/1001/commits/dd8b04a63f301a9cdb2a78a4f0c001e0ae784eeb

packages/card/resource-card/index.tsx
EnMod: Ah, another good immutability catch 🚀 Addressed here:
https://github.com/hashicorp/react-components/pull/1001/commits/dd8b04a63f301a9cdb2a78a4f0c001e0ae784eeb

nandereck (approved):
LGTM!

PR #1005, opened 2023-11-22
fix: correctly determine if array is empty

dstaley:
🎟️ [Asana Task]()

---

## Description

This PR fixes an empty array comparison. In JavaScript, arrays are objects, and equality comparisons only return true if both sides of the operator refer to the same object. By comparing to an empty array literal, this comparison will always return false. In recent versions of TypeScript, this is considered an error, which prevents builds using this code. This PR changes to compare by length instead.

### PR Checklist 🚀

### DIFF
--- a/packages/docs-page/server/get-paths-from-nav-data.ts
+++ b/packages/docs-page/server/get-paths-from-nav-data.ts
@@ -30,7 +30,7 @@ export function getPathsFromNavData(
   const pagePathArrays = getPathArraysFromNodes(navDataResolved)
   // Ensure we include an empty array for the "/" index page path
   // (may be included in nav-data, eg for Terraform, or may not, eg for all other sites)
-  const hasIndexPage = pagePathArrays.filter((p) => p == []).length > 0
+  const hasIndexPage = pagePathArrays.filter((p) => p.length === 0).length > 0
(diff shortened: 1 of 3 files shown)


packages/docs-page/server/get-paths-from-nav-data.ts
```diff
@@ -30,7 +30,7 @@ export function getPathsFromNavData(
   const pagePathArrays = getPathArraysFromNodes(navDataResolved)
   // Ensure we include an empty array for the "/" index page path
   // (may be included in nav-data, eg for Terraform, or may not, eg for all other sites)
-  const hasIndexPage = pagePathArrays.filter((p) => p == []).length > 0
+  const hasIndexPage = pagePathArrays.filter((p) => p.length === 0).length > 0
```
heatlikeheatwave, line 33: The two `.length` methods make this a be confusing to read. What do you think about changing it to this:
```suggestion
  const hasIndexPage = pagePathArrays.some((p) => p.length === 0)
```

dstaley: @heatlikeheatwave love it! Plus it exits early as well, which is faster!

heatlikeheatwave (approved):
🚀 LGTM!

PR #1007, opened 2023-12-06
feat: use next/image instead of react-inline-svg

dstaley:
🎟️ [Asana Task]()

---

## Description

This PR replaces instances of `?include` and `react-inline-svg` with `next/image` or actual inlined SVGs where appropriate. The packages updated in this PR are used on hashicorp.com; other packages likely won't be updated unless they're transitive dependencies of the packages modified in this PR. There may be slight rendering differences due to the switch but this is acceptable given that these components are relatively ancient and should be replaced (all but code-block),

Packages updated:
- '@hashicorp/react-accordion'
	- Chevron on right side
- '@hashicorp/react-alert-banner'
	- Close icon
- '@hashicorp/react-code-block'
	- File icon when filename is rendered
- '@hashicorp/react-logo-grid'
	- Close icon in popup
- '@hashicorp/react-split-rich-cta-list'
	- Arrow on right side
- '@hashicorp/react-tabs'
	- Previous/Next arrows when tabs overflow

### PR Checklist 🚀


packages/accordion/partials/AccordionItems/style.module.css
```diff
@@ -60,7 +60,7 @@
   transition: transform 0.4s;
   transform: var(--icon-transform);
 
-  & svg {
+  & img {
```
pbortnick, line 63: Since this is no longer an svg, the `fill` and `stroke` styles below aren't applied and the color is changed. I know you mentioned we are accepting slight rendering differences in the PR description so if this is one of those disregard!

pbortnick: Honestly, I think this is fine. We are deprecating these components in favor of the hds package now anyways.

pbortnick (approved):
Nice!

PR #1021, opened 2024-04-18
feat: alert banner styling updates

LeahMarieBush:
🎟️ [Asana Task](https://app.asana.com/0/1207142155269495/1207251771790307/f)
🔍 [Preview Link](https://react-components-git-leah-featalert-banner-styling-hashicorp.vercel.app/components/alertbanner)

---

## Description

This PR updates the styling of the alert banner. As part of these changes, there is no longer a close button on the alert so that code has been removed.

### PR Checklist 🚀


LeahMarieBush:
Alert banner styling will remain the same for now

LeahMarieBush:
> Alert banner styling will remain the same for now

After a discussion in Asana, the alert banner changes will be included for the nav updates

packages/alert-banner/style.module.css
```diff
@@ -4,6 +4,9 @@
  */
 
 .root {
+  --token-color-foreground-primary: #3b3d45;
+  --token-color-foreground-faint: #656a76;
```
nandereck, line 8: Are we declaring these variables because we don't have access to the tokens in `react-components`? 

Either way, I wonder if we should remove this, reference the tokens in the declarations below with a fallback to ☝️ hex?

```
color: var(--token-color-foreground-faint, #656A76);
```
This would remove duplication, ensure a token is being used and then, in cases like `react-components` where we don't have the tokens, display the fallback. Thoughts?

packages/alert-banner/index.tsx
```diff
@@ -43,41 +39,23 @@ function AlertBanner({
...
   /**
-   * On mount, hide the banner if it is expired or
-   * if a cookie indicates it was previously closed
+   * On mount, hide the banner if it is expired
    */
   useEffect(() => {
-    const hasBeenDismissed = cookie.get(dismissalCookieId)
     const hasExpired = expirationDate && Date.now() > Date.parse(expirationDate)
-    const shouldBeShown = !hasBeenDismissed && !hasExpired
+    const shouldBeShown = !hasExpired
```
nandereck, line 54: Another non-blocking nit: I could be wrong but, looking at the logic now, I _think_ we could remove this const and just use `hasExpired` in `setIsShown`. Again, totally non-blocking!

nandereck (changes requested):
Just one small comment about the color token. Otherwise, looks awesome 👍

packages/alert-banner/style.module.css
LeahMarieBush: I like this approach much better, thanks for the idea!

nandereck (approved):
Awesome, LGTM!

PR #1038, opened 2025-04-28
chore: audit io components

LeahMarieBush:
🎟️ [Asana Task](https://app.asana.com/1/90955849329269/project/1208440225519245/task/1209887029335163?focus=true)
🔍 [Preview Link](https://react-components-git-leah-choreaudit-io-components-hashicorp.vercel.app/)

---

## Description

This PR removes components that are no longer used as a part of the io component audit. The hero, min-100-layout, and vertical-text-block-list are not in use due to the io site shutdown. There were also failing tests for the consent-manager component due to a failed fetch to the service. Since this consent manager is no longer used (new one [here](https://github.com/hashicorp/web/tree/main/packages/consent-manager)), I decided to remove the tests instead of spending time to fix them.

### PR Checklist 🚀


EnMod (changes requested):
For some reason I can't kick off a canary release on this PR by changing its label? Thinking we should do that before merge, so we can test on `www` and see if any of these components' deps will affect the build. 

Perhaps we have to apply the `release:canary` label to this PR, then add a changeset (which is missing, though not sure if it is needed?) so the PR starts a build?

LeahMarieBush:
@EnMod In the web repo, none of these packages are currently in any of the package.json/package-lock.json files. I don't think that cutting a canary release would make a difference in this case since there would not be anywhere for me to put the new version. I also think the canary release wouldn't work because you can't have a new version of a package that does not exist anymore. I've been looking into previous cases where packages were removed from this repo to see what they did in that case for a canary release and I can't find any examples of this. If you know of any let me know and I will look into them more.

Also since there are not instances where a package was deleted from this repo, it makes me think that maybe deleting these packages is not the right thing to do. I did a search through the hashicorp org and found that there are other packages from this repo that are no longer used but still have the packages available (ex. react-aside, react-call-to-action, react-card, react-related-content, etc). Maybe instead of removing these packages we will instead deprecate this repo at some point when we are no longer using any of the components? Let me know what your thoughts are on all of this!

EnMod:
> none of these packages are currently in any of the package.json/package-lock.json files

Agreed! My main concern is not necessarily with the packages themselves being used, but what dependencies they will remove once they are no longer available. Although yeah, that's probably a moot point if the components aren't being used as deps of other components or being used outright 😅

> since there are not instances where a package was deleted from this repo, it makes me think that maybe deleting these packages is not the right thing to do

Ah that's a good point. Deprecate and stop usage vs delete. I like the idea so we don't crash a test or build somewhere haha. 

> Maybe instead of removing these packages we will instead deprecate this repo at some point when we are no longer using any of the components?

This would be the logical next step yeah IMO, and is a very safe way to go about sunsetting this repo. I'm down for this approach!

EnMod (approved):
Approving in light of our discussion above! Approval is to get this PR out of its Requested Changes state that I placed it in and express support for your solution :+1: Feel free to close the PR!

LeahMarieBush:
Closing this PR due to conversation above!

